PR #1, opened 2023-06-22
chore: add readme, package.json, and probot settings

straker:


.github/settings.yml
```diff
@@ -0,0 +1,36 @@
+repository:
```
straker, line 1: This is mostly taken from our template repo file with the changes of no labels and using `teams` instead of collaborators (which properly configures the repo settings). I found out that if you set labels then it wipes _all_ labels from the repo and only allows those that are set in the probot settings file (not ideal).

AdnoC (approved):
I dislike having probot's settings being called `settings.yml`, but if you can't change it then looks good to me

package.json
```diff
@@ -0,0 +1,18 @@
+{
```
stephenmathieson, line 1: Why is a `package.json` file needed?

PR #13, opened 2023-08-31
chore: add typescript, prettier, eslint, and other standards

straker:
Adding everything that was missing so we can start adding code files and not just `action.yml` files.


michael-siek (approved):

michael-siek:
- [x] reviewed for security

PR #16, opened 2023-08-31
feat(semantic-pr-footer): verify pr footers follow policy

straker:


.github/actions/semantic-pr-footer-v1/action.yml
```diff
@@ -0,0 +1,5 @@
+name: 'Semantic PR Footer'
+description: 'Validate pull request footer against Conventional Commits'
+runs:
+  using: 'node16'
```
michael-siek, line 4:
```suggestion
  using: 'node18'
```

michael-siek (changes requested):

.github/actions/semantic-pr-footer-v1/action.yml
michael-siek: We should use LTS at minimum here

michael-siek (approved):

PR #23, opened 2023-09-06
feat: Add is-release-week github action

dfreedmanpro:
Refs: https://github.com/dequelabs/axe-api-team/issues/372


.github/actions/is-release-week-v1/src/utils.ts
```diff
@@ -0,0 +1,17 @@
+export function getWeekNumber(date: Date): number {
+  const firstDayOfYear: Date = new Date(date.getTime())
+  firstDayOfYear.setMonth(0, 1)
+  const dayOfYear =
+    Math.floor(
+      (date.getTime() - firstDayOfYear.getTime()) / (24 * 60 * 60 * 1000)
+    ) + 1
+  const weekNumber = Math.ceil((firstDayOfYear.getDay() + dayOfYear) / 7)
+
+  return weekNumber
```
straker, line 10: This algorithm isn't quite correct. When asked what the week number is for [Jan 1, 1863](https://luirig.altervista.org/calendar/index.php?year=1863), it returns `0`, and for Jan 4 it returns `1` (should be the 2nd week of the year). Looks like it just counts the days in between the first day of the year and the current date, but doesn't factor in how far into the week the first of the year started.

[This algorithm](https://javascript.plainenglish.io/how-to-get-the-week-of-year-of-a-given-date-in-javascript-c8fe0d764b5d) seems to produce the correct output (testing it further March 30 and April 2 are both week `14`).

```suggestion
  const now = new Date(2021, 3, 1);
  const onejan = new Date(date.getFullYear(), 0, 1);
  const week = Math.ceil((((date.getTime() - onejan.getTime()) / 
  86400000) + onejan.getDay() + 1) / 7);

  return week
```

.github/actions/is-release-week-v1/src/utils.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    })
+  })
+
+  describe('isReleaseWeek()', () => {
+    describe('when current week is odd week', () => {
+      it('returns true if oddWeek is true', () => {
+        assert.isTrue(isReleaseWeek(35, true))
+      })
+
+      it('returns true if oddWeek is false', () => {
```
straker, line 20:
```suggestion
      it('returns false if oddWeek is false', () => {
```

.github/actions/is-release-week-v1/src/utils.test.ts
```diff
@@ -0,0 +1,35 @@
...
+        assert.isFalse(isReleaseWeek(35, false))
+      })
+    })
+
+    describe('when current week is even week', () => {
+      it('returns true if oddWeek is true', () => {
```
straker, line 26:
```suggestion
      it('returns false if oddWeek is true', () => {
```

.github/actions/is-release-week-v1/src/index.ts
```diff
@@ -0,0 +1,16 @@
+import core from '@actions/core'
+import { getWeekNumber, isReleaseWeek } from './utils'
+
+async function main() {
+  try {
+    const oddWeek = core.getInput('oddWeek').toLocaleLowerCase() === 'true'
```
Zidious, line 6: Need to fail the action if `oddWeek` was not supplied: 

```suggestion
    const oddWeek = core.getInput('oddWeek', { required: true }).toLowerCase()   
```

To be safe we should also check the value is what we expect: 

```ts
const oddWeek = core.getInput('oddWeek', { required: true }).toLowerCase()

if (oddWeek !== 'true' || oddWeek !== 'false') { 
	core.setFailed('`oddWeek` must be "true" or "false"')
	return
}

const isOddWeek = oddWeek === 'true'

// ... 

```

.github/actions/is-release-week-v1/action.yml
```diff
@@ -0,0 +1,12 @@
+name: 'Check Release Week'
+description: 'Determine if release should continue to run or stop'
+inputs:
+  oddWeek:
+    description: 'True if the action should run on odd weeks, False if the action should run on even'
+    required: true
```
Zidious, line 6: We can default the value to `false` here: 

```suggestion
  oddWeek:
    description: 'True if the action should run on odd weeks, False if the action should run on even'
    default: 'false'
    required: true
```

.github/actions/is-release-week-v1/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../../tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "."
+  },
```
Zidious, line 5:
```suggestion

```

This can be removed the [root confg](https://github.com/dequelabs/axe-api-team-public/blob/main/tsconfig.json#L8) points to `src`. This is currently causing your `dist` to look like: 

```
dist
	index.d.ts
	utils.d.ts
	index.js
	licenses.txt
src
	index.d.ts
	utils.d.ts
```

`ncc` builds everything into 1 JS file so we should only see this in `dist`: 
```
dist
	index.d.ts
	utils.d.ts
	index.js
	licenses.txt
```

.github/actions/is-release-week-v1/README.md
```diff
@@ -0,0 +1,23 @@
+# is-release-week-v1
+
+A GitHub Action to determine if release should continue to run or stop.
+
+## Inputs
+
+### `oddWeek`
+
+**Required** `true` if the action should run on odd weeks, `false` if the action should run on even
```
Zidious, line 9: nit - feel free to ignore I personally like when the inputs: 
```suggestion
| Name   | Description                                                               | Required | Default |
| ------ | ------------------------------------------------------------------------- | -------- | ------- |
| `oddWeek` | Determine if the action should run on odd weeks of the year | true    | false      |
```

straker (changes requested):
After looking at how `index` isn't getting tests, I'm wondering if we should do something to fix that? Will bring it up at standup

.github/actions/is-release-week-v1/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Outputs
+
+### `isReleaseWeek`
+
+`true` if the week of year matches the `oddWeek` input, `false` otherwise
```
Zidious, line 15: nit - same with outputs ( feel free to ignore too): 

```suggestion
| Name   | Description                                                               | Default |
| ------ | ------------------------------------------------------------------------- | ------- |
| `isReleaseWeek` | `true` if the week of year matches the `oddWeek` input  | false      |
```

Zidious (changes requested):
☕ Small comments inline -

Zidious (commented):
We also need to set the `is-release-week-v1` tests to run within `workflows/test.yml`: 

```yml
  is-release-week-v1:
    runs-on: ubuntu-latest
    timeout-minutes: 5
    needs: lint
    steps:
      - uses: actions/checkout@v3
      - uses: actions/setup-node@v3
        with:
          node-version: 20
          cache: 'npm'
      - run: npm ci
      - run: npm run coverage --workspace=is-release-week-v1
```

.github/actions/is-release-week-v1/src/utils.ts
dfreedmanpro: The current algorithm is the same as the [the one](https://javascript.plainenglish.io/how-to-get-the-week-of-year-of-a-given-date-in-javascript-c8fe0d764b5d) you referred.

It returns as following:
```
  assert.isTrue(getWeekNumber(new Date(1863, 0, 1)) === 1)
  assert.isTrue(getWeekNumber(new Date(1863, 0, 4)) === 2)
```

straker: Oh, you're right. Turns out I was using a different algorithm thinking it was yours. :P

straker:
Here's how we'll start testing the main runner file for github actions https://github.com/dequelabs/axe-api-team-public/pull/24

dfreedmanpro:
@straker Added test for the action run() function.
@Zidious Added is-release-week-v1 action to the `tests` workflow.

.github/actions/is-release-week-v1/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+| Name      | Required | Description                                                                            | Default |
+| --------- | -------- | -------------------------------------------------------------------------------------- | ------- |
+| `oddWeek` | yes      | `true` if the action should run on odd weeks, `false` if the action should run on even | `false` |
+
+## Outputs
+
+| Name            | Description                                            | Default |
+| --------------- | ------------------------------------------------------ | ------- |
+| `isReleaseWeek` | `true` if the week of year matches the `oddWeek` input | false   |
```
Zidious, line 15:
```suggestion
| `isReleaseWeek` | `true` if the week of year matches the `oddWeek` input | `false`  |
```

.github/actions/is-release-week-v1/src/run.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    const core = {
+      getInput: sinon.stub(),
+      setFailed: sinon.spy()
+    }
+
+    run(core as unknown as Core)
+    assert.isTrue(
+      core.setFailed.calledOnceWith('`oddWeek` must be "true" or "false"')
+    )
+  })
```
Zidious, line 20: Oh interesting, so the `action.yml` adds a default value for `oddWeek` to be `false` but because we're testing it directly it doesn't get passed here.

(see comment below) if no input is given it should through the action before it reaches your is true or false statement check. I would've thought the `catch` would get the thrown error from the `{ required: true}` in your `getInput`

straker: The test stubs out `getInput` so it never called into the original `getInput` code to generate the throw.

.github/actions/is-release-week-v1/src/run.test.ts
```diff
@@ -0,0 +1,98 @@
...
+      clock.restore()
+    })
+
+    describe('when it is the odd week', () => {
+      beforeEach(() => {
+        // 2023-09-01, Week #35
+        clock = sinon.useFakeTimers(new Date(2023, 8, 1))
+      })
+
+      it('sets `isReleaseWeek` output correctly based on `oddWeek` input', () => {
```
straker, line 64: These should be split out into two different tests and [should follow single use-case testing](https://github.com/dequelabs/axe-api-team/blob/main/policies/unit-testing.md#single-use-case-testing)

.github/actions/is-release-week-v1/src/run.test.ts
straker: If we want to test that we have the `required: true` added to the getInput call we could ensure that it was called with it:

```js
// don't know for certain if this is works as written, but should get the point across
assert.isTrue(core.getInput.calledWith('oddWeek, sinon.match({ required: true }));
```

dfreedmanpro:
@straker @Zidious Updated the tests

.github/actions/is-release-week-v1/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+| Name      | Required | Description                                                                            | Default |
+| --------- | -------- | -------------------------------------------------------------------------------------- | ------- |
+| `oddWeek` | yes      | `true` if the action should run on odd weeks, `false` if the action should run on even | `false` |
+
+## Outputs
+
+| Name            | Description                                            | Default |
+| --------------- | ------------------------------------------------------ | ------- |
+| `isReleaseWeek` | `true` if the week of year matches the `oddWeek` input | `false` |
```
Zidious, line 15: Saw Steve comment this, I missed it outputs have no defaults its based from the input so this can be - 
```suggestion
| Name            | Description                                           
| --------------- | ------------------------------------------------------ |
| `isReleaseWeek` | `true` if the week of year matches the `oddWeek` input |
```

.github/actions/is-release-week-v1/src/run.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  afterEach(() => {
+    sinon.restore()
+  })
+
+  it('fails if oddWeek is not given', () => {
+    const core = {
+      getInput: sinon
+        .stub()
+        .withArgs('oddWeek', { requied: true })
```
Zidious, line 35:
```suggestion
        .withArgs('oddWeek', { required: true })
```

Zidious (changes requested):
two very minor things otherwise it's LGTM

Zidious (approved):
LGTM

straker (approved):

PR #24, opened 2023-09-07
test: test the main action runner

straker:
Figured out a good way to test our action runner code, taking inspiration from https://github.com/actions/typescript-action/blob/main/__tests__/index.test.ts and improving it with help.


Zidious (approved):
LGTM

PR #25, opened 2023-09-11
feat(.github/actions): add commit list generator

Zidious:
This PR adds the GHA for generating a list of commits between two branches. 

Tested on axe-core-npm: https://github.com/dequelabs/axe-core-npm/actions/runs/6165047966. Please see attached artifact of outputted JSON 

Coverage: https://github.com/dequelabs/axe-api-team-public/actions/runs/6163864589/job/16728350349?pr=25#step:5:80

Closes: https://github.com/dequelabs/axe-api-team/issues/371


tsconfig.json
```diff
@@ -5,7 +5,6 @@
     "lib": ["dom", "es2020"],
     "declaration": true,
     "outDir": "dist",
-    "rootDir": "src",
```
Zidious, line 8: Was causing intellisense to shout when importing files from `src`. This doesn't affect building what we are outputting to `dist`

straker (changes requested):

.github/actions/generate-commit-list-v1/src/getParsedCommitList.ts
```diff
@@ -0,0 +1,42 @@
...
+     * 2. Commit message (e.g. "feat: add new feature")
+     * 3. Pull request ID (e.g. #1234)
+     */
+    const regex = /([a-z0-9]{1,8}) (.*) \((#[0-9]*)\)/
+    const commitMatches = commit.match(regex)
+    //eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const [, sha, title, id] = commitMatches!
+    const idParsed = id.replace('#', '')
+    const type = getCommitType(title)
+    const link = `${repositoryURL}/pull/${idParsed}`
```
straker, line 29: This assumes the commit has the pr number in the title (not always a guarantee). For commits without the pr number in the title you can [use the SHA to get the number using `git ls-remote`](https://stackoverflow.com/a/48085242/2124254)

```
git ls-remote origin 'pull/*/head' | grep <SHA> |  awk -F'/' '{print $3}'
```

Alternatively instead of `awk` you could use `cut` or any other methods to get the pull request number out of the string

.github/actions/generate-commit-list-v1/src/getRepositoryURL.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Gets the URL of the current git repository.
+ * @returns A promise that resolves to a string containing the URL of the current git repository.
+ * @throws An error if the repository URL cannot be retrieved.
+ */
+
+export default async function getRepositoryURL(): Promise<string> {
+  try {
+    const { stdout: repositoryURL } = await getExecOutput('git', [
```
straker, line 11: Rather than doing this we should just require passing the repo in from the github action which already has this info available.

.github/actions/generate-commit-list-v1/src/run.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    setFailed = sinon.spy()
+    setOutput = sinon.spy()
+    getInput = sinon.stub()
+    info = sinon.spy()
+
+    core = {
+      setFailed,
+      setOutput,
+      getInput,
+      info
```
straker, line 31: One of the issues with spying/stubbing all properties is that you don't know if your code is calling things you didn't expect (e.g. you forgot a `return` to stop the code execution).

straker (changes requested):

.github/actions/generate-commit-list-v1/src/test-utils.ts
```diff
@@ -0,0 +1,97 @@
...
+    sha: '1d6220e',
+    type: 'ci',
+    id: '123',
+    link: `${expectedRepositoryURL}/pull/123`
+  },
+  {
+    commit: '5d6220e BREAKING CHANGE: some breaking change (#124)',
+    title: 'BREAKING CHANGE: some breaking change',
+    sha: '5d6220e',
+    type: 'BREAKING CHANGE',
```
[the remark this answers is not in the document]
Zidious: I read this wrong: 

> BREAKING CHANGE: a commit that has a footer BREAKING CHANGE 

_footer_ 🤦

stephenmathieson (commented):
35 files and 5.3k LOCs could be wrapped up nicely in this 1-liner:

```sh
git rev-list --ancestry-path HEAD_OF_BRANCH_A..HEAD_OF_BRANCH_B
```

Have we made this **A LOT** more complicated than necessary?

straker:
@stephenmathieson This parses the commits for us into properties that are needed later (such as checking for breaking changes, if the commits contain a releasable commit, etc.). How we get the commit list from github doesn't so much matter so long as we output a format we can use to parse the data.

.github/actions/generate-commit-list-v1/src/getParsedCommitList.ts
Zidious: This is not going to work, this pulls the list of commits using the _last_ commit SHA of that pull request not the merging commit SHA.

.github/actions/generate-commit-list-v1/src/getParsedCommitList.ts
```diff
@@ -0,0 +1,37 @@
...
+  const parsedCommits: Array<ParsedCommitList> = []
+
+  for (const commit of rawCommitList) {
+    // group sha and title together
+    //eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const shaAndTitle = commit.match(/^(.{7}) (.+?)(?:\s\(#\d+\))?$/)!
+    const sha = shaAndTitle[1]
+    const title = shaAndTitle[2]!
+    const type = getCommitType(title)
+    const id = commit.match(/#(\d+)/)?.[0].replace('#', '') || null
```
straker, line 23: We'll want to make sure this is only grabbing the id from the end and not any random `#number` from the title. Also since you're using a group for the numbers you can use that and not have to replace the `#`

```suggestion
    const id = commit.match(/\(#(\d+)\)$/)?.[1] || null
```

.github/actions/generate-commit-list-v1/src/getParsedCommitList.ts
```diff
@@ -0,0 +1,37 @@
...
+export default function getParsedCommitList({
+  rawCommitList,
+  repository
+}: GetParsedCommitListParams): Array<ParsedCommitList> {
+  const parsedCommits: Array<ParsedCommitList> = []
+
+  for (const commit of rawCommitList) {
+    // group sha and title together
+    //eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const shaAndTitle = commit.match(/^(.{7}) (.+?)(?:\s\(#\d+\))?$/)!
```
straker, line 19: This assumes that the sha is only 7 characters long, which is not always the case. [Git docs](https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#Short-SHA-1) says it grows as the repo commits do 

> Generally, eight to ten characters are more than enough to be unique within a project. For example, as of February 2019, the Linux kernel (which is a fairly sizable project) has over 875,000 commits and almost seven million objects in its object database, with no two objects whose SHA-1s are identical in the first 12 characters.

We'll also probably want to scope the sha to valid hex digits rather than accepting everything

```suggestion
    const shaAndTitle = commit.match(/^([0-9a-f]+) (.+?)(?:\s\(#\d+\))?$/)!
```

Also some additional test cases we'll want:

- `1c41b7a1 revert: feat: awesome feature (#3957) (#3988)` 
- `15a29139 chore(release): 4.8.1`

.github/actions/generate-commit-list-v1/src/getParsedCommitList.ts
straker: That's OK. I think we can try it as a fallback and if it doesn't find anything we just default to `null`. It should help fill in quite a few missing ids though

.github/actions/generate-commit-list-v1/src/run.ts
```diff
@@ -0,0 +1,43 @@
...
+    core.info(`Getting raw commits between ${base} and ${head}...`)
+
+    const rawCommitList = await getRawCommitList({ base, head })
+    const parsedCommitList = getParsedCommitList({
+      rawCommitList,
+      repository: `${github.context.repo.owner}/${github.context.repo.repo}`
+    })
+
+    core.info(`Found ${parsedCommitList.length} commits.`)
+    core.info('Setting output...')
```
straker, line 38: Should log the full list (or at least the titles) for debugging purposes

```suggestion
    core.info(`Found ${parsedCommitList.length} commits.`)
    core.info(JSON.stringify(parsedCommitList, null, 2))
```

straker (dismissed):

straker (approved):

PR #32, opened 2023-09-20
feat(.github/actions): add `create-project-issue-v1` action

Zidious:
This PR does the following: 

- Creates a composite action which combines the [`dabcd/create-issue-action`](https://github.com/dacbd/create-issue-action) pinned at the tagged release commit and the [add-to-board-v1](https://github.com/dequelabs/axe-api-team-public/tree/main/.github/actions/add-to-board-v1) action to create the issue and add it to the board.

Example workflow: https://github.com/dequelabs/zidious-testing/actions/runs/6352697804/job/17256055166#step:3:1

Closes: https://github.com/dequelabs/axe-api-team/issues/370


.github/actions/create-project-issue-v1/src/run.test.ts
```diff
@@ -0,0 +1,308 @@
...
+    info = sinon.spy()
+    getInput = sinon.stub()
+    getExecOutput = sinon.stub(exec, 'getExecOutput')
+  })
+
+  afterEach(sinon.restore)
+
+  const createAndMoveIssueStub = async () => {
+    //`getProjectBoardID` stub
+    getExecOutput.onFirstCall().resolves({
```
straker, line 51: This relies on the order of events which isn't robust. Instead you can use `stub.withArgs` and [sinon matchers](https://sinonjs.org/releases/latest/matchers/) to match the specific calls to their responses. By specifying more detailed matches you can even give different responses based on input.

```suggestion
    getExecOutput.withArgs(sinon.match('gh project view')).resolves({
```

.github/actions/create-project-issue-v1/src/run.test.ts
```diff
@@ -0,0 +1,308 @@
...
+      stdout: JSON.stringify(MOCK_PROJECT_BOARD_ID),
+      exitCode: 0
+    })
+
+    //`addIssueToBoard` stub
+    getExecOutput.onSecondCall().resolves({
```
straker, line 57:
```suggestion
    getExecOutput.withArgs(sinon.match('gh project item-add')).resolves({
```

.github/actions/create-project-issue-v1/src/run.test.ts
```diff
@@ -0,0 +1,308 @@
...
+      stdout: JSON.stringify(MOCK_ISSUE_ADDED),
+      exitCode: 0
+    })
+
+    //`getProjectFieldList` stub
+    getExecOutput.onThirdCall().resolves({
```
straker, line 63:
```suggestion
    getExecOutput.withArgs(sinon.match('gh project field-list')).resolves({
```

.github/actions/create-project-issue-v1/src/run.test.ts
```diff
@@ -0,0 +1,308 @@
...
+      stdout: JSON.stringify(MOCK_FIELD_LIST),
+      exitCode: 0
+    })
+
+    //`moveIssueToColumn` stub
+    getExecOutput.onCall(3).resolves({
```
straker, line 69:
```suggestion
    getExecOutput.withArgs(sinon.match('gh project item-edit')).resolves({
```

.github/actions/create-project-issue-v1/src/run.test.ts
```diff
@@ -0,0 +1,308 @@
...
+      )
+    })
+  })
+
+  describe('when the `project_number` input is not a number', () => {
+    it('throws an error', async () => {
+      getInput.withArgs('github_token', { required: true }).returns('token')
+      getInput.withArgs('title', { required: true }).returns('title')
+      getInput.withArgs('body', { required: true }).returns('body')
+      getInput.withArgs('repository').returns('dequelabs/repo')
```
straker, line 140: These tests aren't very robust as they just ensure you got to the end but not that any of the provided inputs got used as expected (due to mocking). Instead I would look at the specific `getExecOutput` calls and make sure the expected inputs got used for the call. And also would check all the parts individually.

So for example:

```js
it('Looks at the project board using the project board id')
  getInput.withArgs('project_number').returns('70')

  assert.isTrue(getExecOutput.calledWith('gh project view 70 --owner dequelabs --format json'))
```

straker (changes requested):

.github/actions/create-project-issue-v1/src/run.test.ts
```diff
@@ -305,4 +322,68 @@ describe('run', () => {
...
+        getInput,
+        setFailed,
+        setOutput,
+        info
+      }
+
+      await run(core as unknown as Core, github as unknown as GitHub)
+
+      // commands that require the project number
+      const [projectBoarArgs, issueAddedArgs, fieldListArgs] =
```
straker, line 342: This typo is everywhere so will only mark it here, but you'll need to fix it in ever file this has been copy pasted

```suggestion
      const [projectBoardArgs, issueAddedArgs, fieldListArgs] =
```

.github/actions/create-project-issue-v1/src/run.test.ts
```diff
@@ -46,30 +46,89 @@ describe('run', () => {
...
+    title?: string
+    body?: string
+    repository?: string
+    labels?: string
+    assignees?: string
+    project_number?: string
+    column_name?: string
+  }
+
+  const generateInputs = (options?: Partial<GenerateInputs>) => {
```
straker, line 115: 👍

.github/actions/create-project-issue-v1/src/run.test.ts
```diff
@@ -0,0 +1,389 @@
...
+      assert.isTrue(setOutput.calledOnce)
+      assert.isTrue(
+        setOutput.calledWith('issue_url', 'https://deque.bizzy.com')
+      )
+    })
+  })
+
+  describe('when the `assignees` input is provided', () => {
+    it('creates an issue with the assignees', async () => {
+      generateInputs({ assignees: 'foo,bar' })
```
straker, line 284: This test doesn't verify the integration that the issue created used the assignees `foo` and `bar`. You'll need to look at the `getOctokit` create issue stub to verify it used the params.

I would like to see integration tests that verify the inputs provided to the `run` action are being used all the way through the code and not just that you got an issue url back. These tests include:

* verifying that the inputs `title`, `body`, `assignees`, `labels`, and `repository` were all used when creating the issue
* verifying that the inputs `column_name` and `project_number` were used when adding the issue to the project board. the mocks sort of verify this but it is not a direct `assert`, meaning that when it fails the error message will not make sense to fix the problem (the thing that will fail will be something like `info.callCount not called 6 times`. This isn't a helpful message to resolve the problem of the mock not being used because the inputs didn't match)

straker (changes requested):

michael-siek (dismissed):

michael-siek (approved):

michael-siek (approved):

PR #38, opened 2023-10-03
build(./github/actions): rebuild dist for label-and-move-released-issues-v1

Zidious:
We should add a prebuild step that removes the existing `dist` before building another in case we don't need the emitted files the new build generated 

Re-tested: https://github.com/dequelabs/zidious-testing/actions/runs/6395202714/job/17358307503


straker (changes requested):
You'll need to install `rimraf` at the root as I don't think any of the packages have it

Issue #57, opened 2023-10-10
tech-debt: Create a GHA that will re-built dist on merged commits to `main` 


Zidious:
Now we have automated dependency updates in this repository, depending what gets updated they may get used within `.github/actions` because of this, we will need to rebuild `dist`. We can create a GHA that can do the following: 

1. Run on `push` to `main`
2. Create branch `build-update-all-dists` 
3. Run `npm --workspaces run build` which will re-build all `dists` within the action directory
4. If there no file changes, exit 0 
5. If there are file changes, commit all with `build(.github/actions): rebuild dists` 
6. Create PR 

This way we will always be up to date

PR #58, opened 2023-10-10
chore: update dependencies

Zidious:
We had a bit to many dependabot PR's, updated all here so next month we hopefully won't have so many PR's. 

- This PR updates dependencies and rebuilds all actions as they rely some of the updated dependencies, this will have to occur manually, but https://github.com/dequelabs/axe-api-team-public/issues/57 is open to address this issue

No QA Required


.github/actions/label-and-move-released-issues-v1/src/run.test.ts
```diff
@@ -270,7 +254,26 @@ describe('run', () => {
             repo: 'repo'
           }
         },
-        getOctokit: () => octokit
+        getOctokit: () => {
```
Zidious, line 257: As of `@actons/github` v6, they changed the way the generate the GET endpoints which completely broke the way we were calling them, they appear to be proxying old + new endpoints with a new `handler` function which attempted to destructure a variable which ended being undefined: 

>       TypeError: Cannot destructure property ‘decorations’ of ‘endpointMethodsMap.get(...).get(...)’ as it is undefined.



Similar, not exact issue: https://github.com/octokit/plugin-rest-endpoint-methods.js/issues/668


I'll be opening a issue in that repo, but for now this a work around

Zidious: Opened an issue to see if this was an intended change as of v6: https://github.com/octokit/plugin-rest-endpoint-methods.js/issues/683

stephenmathieson:
I do not understand why this repository contains 26+ files. This seems like a huge amount of work to maintain

Zidious:
> I do not understand why this repository contains 26+ files. This seems like a huge amount of work to maintain

As some of the dependencies get used in the other actions, I've had to manually rebuild the `dist` for each action. This should be addressed in https://github.com/dequelabs/axe-api-team-public/issues/57 which will automate everything. In this I am manually updating deps, this is automated also (when it runs on the 1st of every month)

.github/actions/label-and-move-released-issues-v1/src/run.test.ts
Zidious: This issue was fixed by GitHub but the fix needs to be upstreamed to `@actions/github` before I can pull in the change. Will wait for that before merging, moving this to draft.

Zidious: I was waiting for the above PR to be merged but it's not moving. 


Will un-draft this PR and get this moving, opened issue to revert the work around once there has been some movement: https://github.com/dequelabs/axe-api-team-public/issues/80

Zidious:
Un-drafting this PR, please see: https://github.com/dequelabs/axe-api-team-public/pull/58#discussion_r1380166517

straker (approved):

PR #59, opened 2023-10-10
feat: add update-axe-core action

straker:
Migrates https://github.com/dequelabs/update-axe-core to here and updates it to use typescript and tests.

Closes: https://github.com/dequelabs/update-axe-core/issues/7
Closes: https://github.com/dequelabs/update-axe-core/issues/8
Closes: https://github.com/dequelabs/update-axe-core/issues/10


.github/actions/update-axe-core-v1/src/getPackageManager.ts
```diff
@@ -0,0 +1,26 @@
...
+  if (await exists(path.join(dirPath, 'yarn.lock'))) {
+    return 'yarn'
+  }
+
+  return undefined
+}
+
+async function exists(filePath: string): Promise<boolean> {
+  try {
+    await fs.stat(filePath);
```
stephenmathieson, line 21: Asynchronously checking if something exists doesn't make sense. It may exist while this function began and could be gone by the time it's done. This is exactly why Node.js removed `fs.exists()`.

.github/actions/update-axe-core-v1/src/types.ts
```diff
@@ -0,0 +1,8 @@
+import type core from '@actions/core'
+
+export type Core = Pick<
+  typeof core,
+  'setOutput' | 'info' | 'setFailed'
+>
+
+export type getPackageManagerReturn = Promise<'npm' | 'yarn' | undefined>
```
stephenmathieson, line 8: Types should be capitalized. Maybe call this `PackageManager`?

stephenmathieson:
This was ~1k lines of Bash (including comments and stuff that could have been abstracted into less). This PR touches more than 10x that. Are we making a good decision?

.github/actions/update-axe-core-v1/src/getPackageManager.test.ts
```diff
@@ -0,0 +1,49 @@
+import 'mocha'
```
stephenmathieson, line 1: Where are these 3rd party libraries coming from? There's nothing specified in the package manifest.

.github/actions/update-axe-core-v1/README.md
```diff
@@ -0,0 +1,32 @@
...
+  schedule:
+    # Run every night at midnight
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
Zidious, line 26:
```suggestion
      - uses: actions/checkout@v4
```

.github/actions/update-axe-core-v1/src/run.ts
```diff
@@ -0,0 +1,104 @@
+import type { Core } from './types'
+import { glob, type GlobOptions } from 'glob'
```
Zidious, line 2:
```suggestion
import { glob } from 'glob'
```

.github/actions/update-axe-core-v1/src/getPackageManager.ts
straker: Node deprecated `fs.exists` because it [didn't follow node's return style](https://nodejs.org/api/fs.html#fsexistssyncpath).

> fs.exists() is deprecated, but fs.existsSync() is not. The callback parameter to fs.exists() accepts parameters that are inconsistent with other Node.js callbacks. fs.existsSync() does not use a callback.

[`fs.exists` docs says to use `fs.stat` instead](https://nodejs.org/api/fs.html#fsexistspath-callback) for this reason.

> Stability: 0 - Deprecated: Use fs.stat() or fs.access() instead.

.github/actions/update-axe-core-v1/src/run.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    for (const filePath of packages) {
+      core.info(`package.json found in ${filePath}`)
+      const pkg = await import(filePath as string)
+
+      if (
+        !pkg.dependencies?.['axe-core'] &&
+        !pkg.devDependencies?.['axe-core']
+      ) {
+        core.info(`no axe-core dependency`)
```
Zidious, line 36: NBD - 

```suggestion
        core.info(`no axe-core dependency found, moving on...`)
```

dbjorge:
> This was ~100 lines of Bash (including comments and stuff that could have been abstracted into less). This PR touches more than 100x that. Are we making a good decision?

I think it's good to question whether the extra complexity is necessary, but I don't think this is a very fair comparison. The bit equivalent to the 100 lines of bash (the implementation, ie, `action.yml`, non-test/types `ts` files) is 143 lines total here; I think adding 50% LoC to the implementation in order to make it feasible to add unit tests and typechecking is a reasonable tradeoff.

.github/actions/update-axe-core-v1/src/getPackageManager.ts
dbjorge: > It may exist while this function began and could be gone by the time it's done.

This is mostly orthogonal to the choice of async vs sync exists function; this is something you need to consider anytime you do separate calls for exists vs use of a file, even if you use a sync exist check.

.github/actions/update-axe-core-v1/src/run.ts
```diff
@@ -0,0 +1,104 @@
...
+        continue;
+      }
+
+      const dirPath = path.dirname(filePath as string)
+      const packageManager = await getPackageManager(dirPath) ?? rootPackageManager ?? 'npm'
```
Zidious, line 41: Do we want to default to `npm` here, instead of handling `undefined`? 

```suggestion
      const packageManager = await getPackageManager(dirPath) ?? rootPackageManager
      
      if (!packageManager) {
      	core.warning('Unsupported package manager, moving on....'): 
      	continue
      }
```

.github/actions/update-axe-core-v1/src/run.ts
```diff
@@ -0,0 +1,104 @@
...
+      installedMajor !== latestMajor ||
+      installedMinor !== latestMinor
+    ) {
+      core.setOutput('commit-type', 'feat')
+    }
+    else {
+      core.setOutput('commit-type', 'fix')
+    }
+  } catch (error) {
+    console.log(error)
```
Zidious, line 101:
```suggestion
```

Zidious (dismissed):
Minor things inline -

.github/actions/update-axe-core-v1/src/run.ts
```diff
@@ -0,0 +1,104 @@
...
+    const packages = await glob('**/package.json', {
+      absolute: true,
+      cwd,
+      ignore: '**/node_modules/**/package.json'
+    })
+    let installedAxeCoreVersion: string | null = null
+
+    for (const filePath of packages) {
+      core.info(`package.json found in ${filePath}`)
+      const pkg = await import(filePath as string)
```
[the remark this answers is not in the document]
straker: I wish it weren't so, but the return type of Glob seems to be `Array<string | Path>` which is just 🤦

.github/actions/update-axe-core-v1/src/run.ts
straker: We could get into the case where a package prevents the `package-lock.json` creation through an `.nvmrc` file. I think we can assume that `npm` is a safe default if we can't detect it from a file.

dbjorge: I agree with Gabe, though I'd actually go even stronger and consider it an immediate fatal error if we hit a case where we can't reliably determine the package manager, not just a warning. I would much rather fail fast than accept any unnecessary risk of a bot making invalid updates with customer impact.

dbjorge (changes requested):

.github/actions/update-axe-core-v1/src/run.ts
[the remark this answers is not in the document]
Zidious: That should only be the case with if you pass `withFileTypes` does it resolve to `Promise<string[] | path[]>`, otherwise the type should resolve with `string[]`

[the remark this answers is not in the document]
straker: Weird. It was totally throwing errors before but now it's not? 🤷

.github/actions/update-axe-core-v1/src/run.ts
```diff
@@ -0,0 +1,104 @@
...
+
+      if (
+        !pkg.dependencies?.['axe-core'] &&
```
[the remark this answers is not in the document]
straker: I'm not sure. @michael-siek do we have any axe-core peer dependencies that we need to update?

.github/actions/update-axe-core-v1/src/run.ts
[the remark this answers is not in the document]
Zidious:
```ts
if (language.isTypeScript && user.steve) {
	console.error('you shall not pass!')
}
```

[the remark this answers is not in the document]
straker: Every time

.github/actions/update-axe-core-v1/src/run.ts
```diff
@@ -0,0 +1,104 @@
...
+import path from 'path'
+import type { getPackageManagerReturn } from './types'
+
+export default async function run(
+  core: Core,
+  getPackageManager: (dirPath: string) => getPackageManagerReturn,
+  cwd?: string
+) {
+  try {
+    const { stdout: latestAxeCoveVersion } = await getExecOutput('npm', ['info', 'axe-core', 'version'])
```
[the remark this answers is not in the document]
straker: The whole function is wrapped in a try/catch, is this enough of a guard or do you mean handling a `stderr` case?

.github/actions/update-axe-core-v1/src/run.ts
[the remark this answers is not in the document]
michael-siek: I don't think we do

.github/actions/update-axe-core-v1/src/run.ts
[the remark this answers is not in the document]
dbjorge: I meant handling non-zero exitCode - my reading of the `getExecOutput` impl is that it doesn't throw in such a case.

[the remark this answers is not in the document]
straker: I see what you mean. That's a good catch and definitely want to handle that.

.github/actions/update-axe-core-v1/src/run.ts
```diff
@@ -0,0 +1,133 @@
...
+      } = await getExecOutput(packageManager, [
+        packageManager === 'npm' ? 'i' : 'add',
+        dependencyType,
+        `axe-core@${pinStrategy}${latestAxeCoreVersion}`
+      ], {
+        cwd: dirPath
+      })
+      if (installError) {
+        throw new Error(`Error installing axe-core:\n${installError}`)
+      }
```
Zidious, line 109:
```suggestion
      if (installExitCode) {
        throw new Error(`Error installing axe-core:\n${installError}`)
      }
```

Zidious (changes requested):
One minor comment inline -

Zidious (approved):

Issue #63, opened 2023-10-13
Make GHA pull latest version


dequejenn:
We need to be able to pull the latest version of a release to use in things like https://github.com/dequelabs/axe-api-team-public/issues/61 and https://github.com/dequelabs/axe-api-team-public/issues/62 . Unsure best way - parse from title? Use latest tag (if all repos use tags)? Read from the latest output of prepare_release.sh?

Scope change per Steve (11/21) which sounds just like https://github.com/dequelabs/axe-api-team-public/issues/82
1. We don't need to get the version number. At any point we would call this we can just look at the version of package.json in the current repo
2. For the docs repo there can be multiple open issues labeled release, either for different repos or even the same repo if multiple versions were released in quick succession and the docs for the prior one haven't been written yet. We'll need to filter all the open release issues by the name of the github repo and also by the current version. Allow supplying version number to input but default to current version from package.json
3. We'll need to allow the action to get issues from other repos (namely the docs repo). Add input for repo (default to current context repo)
4. Rename action to get-release-issue

dequejenn:
Closing because this scope was for `get-version-number` which is no longer needed. Only https://github.com/dequelabs/axe-api-team-public/issues/82 is needed.

PR #64, opened 2023-10-13
feat(.github/actions): add `get-issue-url-and-version-number-v1` action

Zidious:
This PR add an action that gets the issue URL And version number from the release-candidate issue and e.g. use for posting slack notifications. Adding this small GHA here means it lives in once place as it will be added to 5+ repositories. 

Example workflow file:https://github.com/dequelabs/zidious-testing/actions/runs/6696030017/workflow
Example workflow: https://github.com/dequelabs/zidious-testing/actions/runs/6696030017/job/18192888009

Closes: https://github.com/dequelabs/axe-api-team-public/issues/63, https://github.com/dequelabs/axe-api-team-public/issues/82


.github/actions/get-version-number-and-issue-url-v1/README.md
```diff
@@ -0,0 +1,43 @@
+# get-version-number-and-issue-url-v1
+
+This action gets the version number and issue url from the release-candidate.
+
+## Inputs
+
+| Input                   | Description                                                   | Required | Default |
+| ----------------------- | ------------------------------------------------------------- | -------- | ------- |
+| `disable-get-issue-url` | Whether or not to get the issue URL for the release candidate | false    | `false` |
```
straker, line 9: If you are providing an input to disable half of this action, it would probably be best to split this action out into two separate actions

.github/actions/get-version-number-and-issue-url-v1/action.yml
```diff
@@ -0,0 +1,51 @@
...
+        issueUrl=$(gh issue list --repo ${{ github.repository }} --label release --state open --json url --jq '.[0].url')
+        echo "Issue URL: $issueUrl"
+
+        echo "issue-url=$issueUrl" >> $GITHUB_OUTPUT
+
+    - name: Get release-candidate version number
+      id: get_version_number
+      shell: bash
+      run: |
+        pullRequestID=$(gh pr list --repo ${{ github.repository }} --json number -s merged --base release --jq '.[0].number')
```
straker, line 41: The issue you get in the `Get release-candidate issue URL` step should also include the version number in the title. It would be easier to get the version number from there rather than making an additional 2 calls to pull requests

straker (dismissed):

straker (changes requested):
I know this is small but I think this should be a typescript file and include tests. `is-release-in-progress` doesn't do too much more but does have tests.

This also doesn't need to be a composite action anymore as it only has the 1 step.

.github/actions/get-issue-url-and-version-number-v1/src/run.ts
```diff
@@ -0,0 +1,44 @@
...
+    core.info(`Found issue: ${JSON.stringify(issue, null, 2)}`)
+
+    const issueUrl = issue.url
+    core.info(`Issue URL: ${issueUrl}`)
+
+    const issueTitle = issue.title
+    core.info(`Issue title: ${issueTitle}`)
+
+    // get version number from issue title (e.g. "1.33.7")
+    const versionNumber = issueTitle.match(/\d+.\d+.\d+/)
```
straker, line 29: Need to escape the periods

```suggestion
    const versionNumber = issueTitle.match(/\d+\.\d+\.\d+/)
```

.github/actions/get-issue-url-and-version-number-v1/src/run.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    if (!versionNumber) {
+      throw new Error(`Could not find version number in issue title`)
+    }
+    core.info(`Version number: ${versionNumber}`)
```
straker, line 34: `String.match` returns an array of matched info so you'll probably want to grab the first index which will be the match

```js
const versionMatch = issueTitle.match(/\d+\.\d+\.\d+/)

if (!versionMatch) {
  throw new Error(`Could not find version number in issue title`)
}

const versionNumber = versionMatch[0]
```

.github/actions/get-issue-url-and-version-number-v1/src/run.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    core.info(
```
straker, line 36: This info's probably not needed since we just output both variables right before

.github/actions/get-issue-url-and-version-number-v1/src/run.test.ts
```diff
@@ -0,0 +1,127 @@
...
+        exitCode: 0
+      })
+
+      await run(core, github)
+
+      assert.isTrue(
+        getExecOutput.calledWith(
+          'gh issue list --repo owner/repo --label release --state open --json url,title'
+        )
+      )
```
straker, line 81: This assert doesn't belong in this test as it's not really related to the failed action and goes against our [single use case testing](https://github.com/dequelabs/axe-api-team/blob/main/policies/unit-testing.md#single-use-case-testing)

.github/actions/get-issue-url-and-version-number-v1/src/run.test.ts
```diff
@@ -0,0 +1,127 @@
...
```
straker, line 120: Ditto for this assert

.github/actions/get-issue-url-and-version-number-v1/src/run.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+      assert.equal(setOutput.callCount, 2)
+      assert.equal(setOutput.getCall(0).lastArg, url)
+      assert.equal(setOutput.getCall(1).lastArg, '1.33.7')
```
straker, line 124: How is this working if `String.match` returns an array?

.github/actions/get-issue-url-and-version-number-v1/src/run.ts
```diff
@@ -0,0 +1,44 @@
...
+import { getExecOutput } from '@actions/exec'
+
+export default async function run(core: Core, github: GitHub) {
+  const { repo, owner } = github.context.repo
+  try {
+    const {
+      stdout: rawIssue,
+      stderr: issueError,
+      exitCode: issueExitCode
+    } = await getExecOutput(
```
straker, line 11: This assumes that this code will always find an issue that matches. We should build in an extra guards for when that isn't the case

.github/actions/get-issue-url-and-version-number-v1/src/run.ts
```diff
@@ -0,0 +1,44 @@
...
+      `gh issue list --repo ${owner}/${repo} --label release --state open --json url,title`
+    )
+
+    if (issueExitCode) {
+      throw new Error(`Error getting issue:\n${issueError}`)
+    }
+
+    const [issue] = JSON.parse(rawIssue.trim()) as Issue[]
```
straker, line 19: This assumes that the `gh issue list` command will always find an issue that matches. We should build in guards just in case it doesn't

straker (changes requested):

.github/actions/get-issue-url-and-version-number-v1/src/run.test.ts
Zidious: 🧙 with `setOutput` see below: 

```ts
import * as core from "@actions/core";

const a = "chore: RC v1.0.0";

const b = a.match(/\d+\.\d+\.\d+/);

if (!b) {
  throw new Error("No version found");
}

// takes the 1st value i.e. 1.0.0
core.info(`Found version: ${b}`);

// @param value — value to store. Non-string values will be converted to a string via JSON.stringify
core.setOutput("version", b);
```

image


We should update it rather than rely on the parsing from the GHA

Zidious:
Re-ran test worflow: https://github.com/dequelabs/zidious-testing/actions/runs/6721071926/job/18265974025

.github/actions/get-issue-url-and-version-number-v1/src/run.test.ts
```diff
@@ -0,0 +1,140 @@
...
+        exitCode: 0
+      })
+
+      await run(core, github)
+
+      assert.equal(setOutput.callCount, 2)
+      assert.equal(setOutput.getCall(0).lastArg, url)
+      assert.equal(setOutput.getCall(1).lastArg, '1.33.7')
+    })
+  })
```
straker, line 139: Should add a failed test for the try/catch

```js
it('fails if anything throws', async () => {
  const core = {
    setFailed,
    info() { throw new Error('failure!') }
  }
  await run(core as unknown as Core, getPackageManagerStub)

  assert.isTrue(core.setFailed.calledWith('failure!'))
})
```

.github/actions/get-issue-url-and-version-number-v1/README.md
```diff
@@ -0,0 +1,37 @@
...
+| ---------------- | ----------------------------------------------------------- |
+| `version-number` | The version number of the release candidate.                |
+| `issue-url`      | The url of the issue associated with the release candidate. |
+
+## Example usage
+
+```yaml
+name: Get version number and issue url and post to Slack
+
+on: push
```
straker, line 17: Should specify that we only want this to run when the release branch is pushed to

```suggestion
on: 
  push:
    branches:
      - release
```

straker (changes requested):

straker:
We are changing how we are doing this. Instead of opening the QA ticket from the create-release-candidate action, we'll be creating a new action that is run on push to `release` branch that will build and deploy the release, then we will create the QA ticket. Since we'll be in the branch that has all the information, we don't need to look for tickets and can just use the version from `package.json` and the current commit sha to put into the ticket.

dequejenn:
We do need this, as we are not creating the issue during this step. We are creating the issue in `create-release-candidate` and then adding a comment to it, moving the issue around, and sending to slack in `deploy-release-candidate`

straker (changes requested):
(see prior comment)

Zidious:
Superseded by: https://github.com/dequelabs/axe-api-team-public/pull/101

Issue #30, opened 2023-09-15
tech-debt:  move the repo specific implementations of updating axe-core to this repo


Zidious:
See comment: https://github.com/dequelabs/axe-api-team-public/pull/26#discussion_r1326120288

straker:
We already have https://github.com/dequelabs/update-axe-core, but we should port it to this repo and update it to use TypeScript, add tests, and resolve all the open issues. We should also create a composite action that uses it and opens a pr for us as updating axe-core without a pr is kinda useless. Plus every repo that uses the action has to duplicate the pr creation part.

PR #66, opened 2023-10-16
feat: add create-update-axe-core-pull-request-v1

straker:
Creates a single action to update axe-core and open a pull request. We investigated using dependabot to do this for us as it maintains the pinning strategy, but dependabot does not allow multiple entries for different ecosystems. So we can't define both a monthly cadence for everything excluding axe-core and a daily cadence for axe-core.

Requires: https://github.com/dequelabs/axe-api-team-public/pull/65

Closes: https://github.com/dequelabs/update-axe-core/issues/9
Closes: https://github.com/dequelabs/axe-api-team-public/issues/30

Successful action run: https://github.com/dequelabs/straker-test-repo/actions/runs/6549902423/job/17787837546
Pr to update axe-core: https://github.com/dequelabs/straker-test-repo/pull/20
Action cancel after up-to-date: https://github.com/dequelabs/straker-test-repo/actions/runs/6549828140/job/17787609809

### DIFF
--- a/.github/actions/update-axe-core-v1/src/run.ts
+++ b/.github/actions/update-axe-core-v1/src/run.ts
@@ -83,7 +83,6 @@ export default async function run(
       }
 
       core.info(`current axe-core version ${axeCoreVersion}`)
-      core.info(`installedAxeCoreVersion before "${installedAxeCoreVersion}"`)
 
       // we expect that all axe-core versions will be the same
       // for every package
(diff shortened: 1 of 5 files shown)


Zidious (approved):

WilcoFiers (commented):
Do we still need this now that Dependabot has gotten better?

straker:
We're only running dependabot monthly so this won't help us update axe-core quickly enough (the week after it releases).

.github/actions/update-axe-core-v1/src/run.ts
```diff
@@ -11,14 +11,16 @@ export default async function run(
...
-      stdout: latestAxeCoreVersion,
+      stdout: npmInfoOut,
       stderr: npmInfoError,
       exitCode: npmExitCode
     } = await getExecOutput('npm', ['info', 'axe-core', 'version'])
     if (npmExitCode) {
       throw new Error(`Error getting latest axe-core version:\n${npmInfoError}`)
     }
 
+    // `npm info`` adds newline character to end of the output
```
straker, line 22: While testing saw that the build still created a pr even when versions were up-to-date. Turns out that the `stdout` of `npm info` had a newline character at the end so the check for equality failed and still would try to install axe-core@latest

Zidious (approved):

PR #95, opened 2023-11-21
fix(.github/actions/semantic-pr-footer-v1): import github properly

michael-siek:
No QA needed


.github/actions/semantic-pr-footer-v1/src/index.ts
```diff
@@ -1,5 +1,5 @@
 import core from '@actions/core'
-import github from '@actions/github'
+import * as github from '@actions/github'
```
Zidious, line 2: We'll need the same import for `core` too otherwise it'll break in a later step

Zidious (changes requested):

Zidious (approved):

Issue #82, opened 2023-11-09
Create public action get-release-issue


dequejenn:
### Overview ###

**Repo:** https://github.com/dequelabs/axe-api-team-public
**GHA Name:** get-release-issue
**Purpose:** To get a reference to the latest open issue in the repo with the label `release`
**Usage:** This GHA will be called from within a parent GHA

**Inputs:**  

None

**Outputs:**
- Reference to the open `release` Github issue

***
### Details ###
This GHA should look for an issue in the repo that is labelled `release` and is currently open. If there are more than 1 returned, it should return the latest. The output will be used in a different GHA to add a comment onto this issue, so the output should be whatever GHA can use to add a comment (ID?, URL?)

The action should also have unit tests and a Readme. See https://github.com/JasonEtco/create-an-issue for template on how to do this.

Each of the public actions in https://github.com/dequelabs/axe-api-team-public should include a readme with information on how to use the action and a list all inputs and outputs.

PR #101, opened 2023-12-04
feat(.github/actions): add `get-release-issue-v1` action

Zidious:
This PR adds a new GHA for getting a issue URL for a given repo and version number. 

- The input `version` is required and is likely going to be part of a composite action to grab the version from package.json (for example) and passed to this action
- The input `owner-and-repo` can be overridden to something like `dequelabs/my-repo` but defaults to current repo
- Added tests accordingly

Internal-testing: 

Given a single issue: https://github.com/dequelabs/zidious-testing/actions/runs/7090495379/job/19297573314
Given multiple issues and repos:  https://github.com/dequelabs/zidious-testing/actions/runs/7090546614/job/19297744680
Given the issue doesn't exist: https://github.com/dequelabs/zidious-testing/actions/runs/7090586377/job/19297887126
Given an unknown `owner-and-repo` input: https://github.com/dequelabs/zidious-testing/actions/runs/7090579202/job/19297860895

Ref: https://github.com/dequelabs/axe-api-team-public/pull/64#issuecomment-1821151278
Closes: https://github.com/dequelabs/axe-api-team-public/issues/82


.github/actions/get-release-issue-v1/src/run.test.ts
```diff
@@ -0,0 +1,310 @@
...
+      assert.isTrue(core.info.calledTwice)
+      assert.isTrue(
+        core.info.secondCall.calledWith(
+          `Found issue: ${issue.url}. Setting "issue-url" output...`
+        )
+      )
+
+      assert.isTrue(setOutput.calledOnce)
+      assert.isTrue(setOutput.calledWith('issue-url', issue.url))
+    })
```
straker, line 118: This is testing multiple things and ensuring that each portion of the code is called in a single test. Please split this type of test out and use [single use case testing](https://github.com/dequelabs/axe-api-team/blob/main/policies/unit-testing.md#single-use-case-testing). Other tests also test multiple outputs in a single test and should be refactored.

.github/actions/get-release-issue-v1/src/run.ts
```diff
@@ -0,0 +1,57 @@
...
+       * @see https://github.com/dequelabs/axe-api-team-public/blob/main/.github/actions/create-release-candidate-v1/action.yml#L170
+       */
+      ({ title }) =>
+        isDocsRepo
+          ? title === `${owner}/${repo} v${version}`
+          : title === `${ownerAndRepo} v${version}`
+    )
+
+    if (!issue) {
+      throw new Error(`No issue found for ${ownerAndRepo} v${version}`)
```
straker, line 49: Do we want this action to fail if it can't find a ticket? What should happen if we manually close a ticket before running this action? Failing would mean we cannot automatically revert the release since a single ticket isn't found that may not need to be closed.

straker (changes requested):

.github/actions/get-release-issue-v1/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+jobs:
+  get-release-issue:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dequelabs/axe-api-team-public/.github/actions/get-release-issue-v1@main
+        id: get-release-issue
+        with:
+          version: '1.0.0'
```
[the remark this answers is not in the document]
Zidious: Yes - IMO this action should only be responsible for getting a release-issue not getting the version and release issue. We may have repos that do not have `lerna.json` and have a `package.json` for the versioning. Example: 

https://github.com/dequelabs/axe-api-team-public/pull/102/files#diff-58d5337330f7e542de2ed43b7f979a91c8ea07477f1397b364e19e9816d5ff48R20-R36

.github/actions/get-release-issue-v1/action.yml
```diff
@@ -0,0 +1,17 @@
+name: 'Get release issue'
+description: 'A GitHub action for getting the release issue for a release candidate'
+
+inputs:
+  version:
+    description: 'The version of the release candidate'
+    required: true
+  owner-and-repo:
```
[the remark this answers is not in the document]
Zidious: Good call!

.github/actions/get-release-issue-v1/src/run.ts
Zidious: Discussed that we should warn the issue instead of erroring here

Zidious:
Test workflows - 

When the issue doesn't exist: https://github.com/dequelabs/zidious-testing/actions/runs/7119956090/job/19386169053
When the issue exists: https://github.com/dequelabs/zidious-testing/actions/runs/7119986925/job/19386264587
When more than one issue exists: https://github.com/dequelabs/zidious-testing/actions/runs/7120000178/job/19386309758
When the repo doesn't exist: https://github.com/dequelabs/zidious-testing/actions/runs/7120009436/job/19386342018

straker (changes requested):
One final test is [the test of the top-level try/catch](https://github.com/dequelabs/axe-api-team-public/blob/main/.github/actions/add-to-board-v1/src/run.test.ts#L374-L391)

.github/actions/get-release-issue-v1/src/run.test.ts
```diff
@@ -245,10 +245,10 @@ describe('run', () => {
       })
     })
 
-    describe('is for the docs repo', () => {
+    describe('when passed owner and', () => {
```
straker, line 248:
```suggestion
    describe('when passed owner and repo', () => {
```

straker (changes requested):
Just a small nit in a test description

straker (approved):

PR #105, opened 2023-12-07
chore: add mocha eslint plugin and block `only` tests

Zidious:
Added mocha' eslint plugin to block `.only` tests from being pushed (see https://github.com/dequelabs/axe-api-team-public/actions/runs/7133577665/job/19426575903?pr=101)

No QA Required

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,11 +6,17 @@ module.exports = {
     'eslint:recommended',
     'plugin:@typescript-eslint/eslint-recommended',
     'plugin:@typescript-eslint/recommended',
+    'plugin:mocha/recommended'
   ],
   rules: {
     '@typescript-eslint/camelcase': 'off',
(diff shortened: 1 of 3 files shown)


WilcoFiers (dismissed):
Do we really need this?

Zidious:
> Do we really need this?

It doesn't harm anything to include it. We shouldn't allow `only` tests from being pushed

straker:
Do we get any additional benefit to using mocha/recommended or is the only reason to use it is to get the `.only` guard? If so we should use `mocha-no-only` instead as it snipes the problem and doesn't add any additional rules we want to turn off

Zidious:
We require the plugin to be installed, otherwise: 

```
  1:1  error  Definition for rule 'mocha/no-exclusive-tests' was not found  mocha/no-exclusive-tests
```

will close for now until it becomes a problem, we can revisit if we need too

Issue #92, opened 2023-11-21
Require QA consideration in PRs


dequejenn:
Set up https://github.com/dequelabs/axe-api-team-public/blob/main/.github/actions/semantic-pr-footer-v1 so it will block a PR that does not have at least one of the following:

- There is a link to a valid issue in the description of the PR
- The words "QA NOTES" appear in the description of the PR
- The words "QA NOT NEEDED" appear in the description of the PR

gayathrit12:
This is working as expected.

Workflow is defined in the below link:

https://github.com/dequelabs/axe-api-team-public/blob/e04018a727e4805bec99f9fdccffd524ac1b4cdc/.github/workflows/semantic-pr-footer.yml#L17

## Screenshot:

![image](https://github.com/dequelabs/axe-api-team-public/assets/102580833/22dce016-8f80-40d4-9063-7ec459565acd)

PR #106, opened 2023-12-08
ci: add timeouts for each GHA workflow and QA pr footer

michael-siek:
Closes: #92 

No QA required


WilcoFiers (changes requested):
Action seems to be busted

Zidious (approved):

Issue #91, opened 2023-11-17
generate-commit-list needs to work off tag and/or HEAD


straker:
After the `release` branch has been merged into `main`/`master`, we want to use [label-and-move-released-issues-v1](https://github.com/dequelabs/axe-api-team-public/tree/main/.github/actions/label-and-move-released-issues-v1) on all commits in the release. However using `generate-commit-list` requires two branches, but once master is merged all branches are in-sync there won't be any differences between the branch commits.

To address this we'll need `generate-commit-list` to accept a tag and the term `HEAD` in order to get a commit list from the last git tag to the current `main`/`master` HEAD.

```
git log v4.8.2..HEAD --oneline --no-merges --abbrev-commit
```

PR #110, opened 2023-12-20
feat(.github/actions): add `tag` input option for generate-commit-list-v1 action

Zidious:
This PR adds the input `tag` to allow the action to generate between a tag and `HEAD` instead of always `base` and `head` branches. 

With tag example workflow: https://github.com/dequelabs/zidious-testing/actions/runs/7281401120/job/19841836837#step:3:61
Workflow file: https://github.com/dequelabs/zidious-testing/actions/runs/7281401120/workflow

Closes: https://github.com/dequelabs/axe-api-team-public/issues/91


.github/actions/generate-commit-list-v1/src/doesBranchOrTagExist.ts
```diff
@@ -0,0 +1,31 @@
+import { getExecOutput } from '@actions/exec'
```
Zidious, line 1: Refactored name so it's a branch name or tag

.github/actions/generate-commit-list-v1/src/run.ts
```diff
@@ -1,34 +1,64 @@
...
 
-    const [doesBaseExist, doesHeadExist] = await Promise.all([
-      doesBranchExist(base),
-      doesBranchExist(head)
-    ])
-
-    core.info(`Checking if ${base} exists...`)
-    if (!doesBaseExist) {
-      core.setFailed(`The base branch ${base} does not exist.`)
+    if (!tag && (!base || !head)) {
```
straker, line 12: Can apply [De Morgan's Law](https://brilliant.org/wiki/de-morgans-laws/) to simplify the statement a bit

```suggestion
    if (!(base && head || tag)) {
```

.github/actions/generate-commit-list-v1/src/run.test.ts
```diff
@@ -49,17 +47,17 @@ describe('run', () => {
...
+        setFailed.calledWith(
+          'You must provide either a tag or both a base and head branch.'
+        )
       )
     })
   })
 
   describe('when the head input is not provided', () => {
     it('should throw an error', async () => {
       getInput.withArgs('base', { required: true }).returns('main')
```
straker, line 59: `required: true` is no longer passed so the stub isn't ever called. This still passed the test though as each of the inputs were missing, but failing for the wrong reason.

```suggestion
      getInput.withArgs('base').returns('main')
```

.github/actions/generate-commit-list-v1/src/run.test.ts
```diff
@@ -49,17 +47,17 @@ describe('run', () => {
...
-      getInput.withArgs('head', { required: true }).throws({
-        message: 'Input required and not supplied: head'
-      })
+      getInput.withArgs('head', { required: true }).returns('')
```
straker, line 60:
```suggestion
      getInput.withArgs('head').returns('')
```

.github/actions/generate-commit-list-v1/src/run.test.ts
```diff
@@ -70,14 +68,63 @@ describe('run', () => {
...
+        assert.isTrue(
+          setFailed.calledWith(
+            'You cannot provide both a tag and both a base and head branch.'
+          )
+        )
+      })
+    })
+
+    describe('and the head input is provided', () => {
+      it('should throw an error', async () => {
```
straker, line 102: Should have another test to verify when both `head` and `base` are provided

.github/actions/generate-commit-list-v1/src/run.test.ts
```diff
@@ -49,17 +47,17 @@ describe('run', () => {
...
       assert.isTrue(
-        setFailed.calledWith('Input required and not supplied: base')
+        setFailed.calledWith(
+          'You must provide either a tag or both a base and head branch.'
+        )
       )
     })
   })
 
   describe('when the head input is not provided', () => {
```
straker, line 57: Should have another test for when the tag input is not provided

.github/actions/generate-commit-list-v1/src/getRawCommitList.ts
```diff
@@ -10,15 +10,14 @@ import { GetRawCommitListParams } from './types'
...
 }: GetRawCommitListParams): Promise<string[]> {
   try {
-    const { stdout: rawCommitList } = await getExecOutput('git log', [
-      `origin/${base}..origin/${head}`,
-      '--oneline',
-      '--no-merges',
-      '--abbrev-commit'
-    ])
+    const subCommand = tag ? `${tag}..HEAD` : `origin/${base}..origin/${head}`
+    const { stdout: rawCommitList } = await getExecOutput(
```
straker, line 18: Should account for `stderr` errors

.github/actions/generate-commit-list-v1/src/getRawCommitList.ts
```diff
@@ -10,15 +10,14 @@ import { GetRawCommitListParams } from './types'
...
+  tag
 }: GetRawCommitListParams): Promise<string[]> {
   try {
-    const { stdout: rawCommitList } = await getExecOutput('git log', [
-      `origin/${base}..origin/${head}`,
-      '--oneline',
-      '--no-merges',
-      '--abbrev-commit'
-    ])
+    const subCommand = tag ? `${tag}..HEAD` : `origin/${base}..origin/${head}`
```
straker, line 17: This change should have tests in `getRawCommitList.test.ts`

.github/actions/generate-commit-list-v1/src/doesBranchOrTagExist.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @returns A promise that resolves to a boolean indicating whether the branch or tag exists.
+ */
+
+export default async function doesBranchOrTagExist({
+  branchName,
+  tag
+}: DoesBranchOrTagExistParams): Promise<boolean> {
+  try {
+    const subCommand = tag ? tag : `origin/${branchName}`
+    await getExecOutput(`git rev-parse --verify ${subCommand}`)
```
straker, line 21: Does this need to account for `stderr` errors?

.github/actions/generate-commit-list-v1/src/doesBranchOrTagExist.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  beforeEach(() => {
+    getExecOutputStub = sinon.stub(exec, 'getExecOutput')
+  })
+
+  afterEach(() => {
+    getExecOutputStub.restore()
+  })
+
+  describe('when the branch exists', () => {
+    it('returns true', async () => {
```
straker, line 19: You should verify that the code is calling the git command with the correct input. Right now you only modify the stub return but if passing a branch actually called the tag code path this test would still pass even though that is incorrect. You'll need to do this for the tag parameter as well.

straker (changes requested):

.github/actions/generate-commit-list-v1/src/run.test.ts
maksym-shynkarenko: we have 3 test cases. Do I have to add something else?

https://github.com/dequelabs/axe-api-team-public/blob/feat-add-tags-commmit-list/.github/actions/generate-commit-list-v1/src/run.test.ts#L124-L242

.github/actions/generate-commit-list-v1/src/run.test.ts
maksym-shynkarenko: I think this test case covers the case when nothing is provided

https://github.com/dequelabs/axe-api-team-public/blob/feat-add-tags-commmit-list/.github/actions/generate-commit-list-v1/src/run.test.ts#L37-L55

.github/actions/generate-commit-list-v1/src/getRawCommitList.ts
maksym-shynkarenko: @straker 
the function `getExecOutput` returns values in `stdout` if everything is ok and throws an error if not. `stderr` is never thrown. I've tested this

.github/actions/generate-commit-list-v1/src/doesBranchOrTagExist.ts
maksym-shynkarenko: the function `getExecOutput` returns values in stdout if everything is ok and throws an error if not. `stderr` is never thrown. I've tested this

.github/actions/generate-commit-list-v1/src/run.test.ts
straker: This still needs to be corrected as the stub is not being called as `{ required: true}` is not passed. This means it's failing because the `base` input is not being set and not because the `head` input.

.github/actions/generate-commit-list-v1/src/getRawCommitList.ts
```diff
@@ -10,15 +10,14 @@ import { GetRawCommitListParams } from './types'
...
+    const { stdout: rawCommitList } = await getExecOutput(
+      `git log ${subCommand} --oneline --no-merges --abbrev-commit`
```
straker, line 19: This works, but worth nothing that the way `getExecOutput` works is that it will escape every character of this command. The command right now shouldn't be affected by escaping but it's worthing nothing for the future if we change the command that escaping would be problematic. In which case we need to go back to the `exect(command, [args])` style.

straker (changes requested):
Looking good, just one minor fix in the test

.github/actions/generate-commit-list-v1/src/getRawCommitList.ts
maksym-shynkarenko: for the future we will use `exec(commandLine: string, args?: string[], options?: ExecOptions)`

PR #117, opened 2024-01-09
fix(.github/actions/update-axe-core-v1): get package manager not just from root level

michael-siek:


straker:
No longer needed as we figured out the problem was trying to cache npm in the setup node action when there wasn't a `package-lock.json` file at the root.

PR #119, opened 2024-01-09
fix(semantic-pr-footer-v1): ignore empty lines at end of body

straker:
If the body of the pr had newlines at the end the action would fail. This fixes that so empty newlines are removed before parsing the last line. The rest of the files changed are done automatically by our pre-commit hook (sorry it's longer that way).

No qa required


.github/actions/semantic-pr-footer-v1/src/run.test.ts
```diff
@@ -136,6 +136,28 @@ describe('run', () => {
     assert.isTrue(core.info.calledWith('Footer matches team policy'))
   })
 
+  it('ignores empty newlines as last line', () => {
```
straker, line 139: Here's the relevant change

.github/actions/semantic-pr-footer-v1/src/run.ts
```diff
@@ -34,7 +34,7 @@ export default function run(core: Core, github: GitHub) {
       return
     }
 
-    const bodyLines = body.split(/[\r\n]+/)
+    const bodyLines = body.trim().split(/[\r\n]+/)
```
straker, line 37: Also relevant change

Zidious (approved):
LGTM

Issue #107, opened 2023-12-13
semantic pr footer should support linking the same way github does


stephenmathieson:
The `semantic-pr-footer` action only supports a _tiny_ subset of the keywords GitHub supports. This seems like an unnecessary limitation, which has already caused confusion (https://github.com/dequelabs/axe-api-team/pull/460).

See https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #40, opened 2023-10-04
feat(.github/actions/create-release-candidate-v1): add create-release-candidate action

Zidious:
This PR adds all the created actions in `.github/actions` required for creating a release candidate. 

Rendered [README](https://github.com/dequelabs/axe-api-team-public/pull/40/files?short_path=104c593#diff-104c593ec158efd56904cc736abd042d6414cbc1579ab324486cc4f68695ed3b)

edit: see below, code updated new examples generated
~~Example workflow file: https://github.com/dequelabs/zidious-testing/blob/develop/.github/workflows/create-release-candidate.yml
Example workflow: https://github.com/dequelabs/zidious-testing/actions/runs/6433162468
Example issue created: https://github.com/dequelabs/zidious-testing/issues/83
Example release candidate created: https://github.com/dequelabs/zidious-testing/pull/82
Example docs issue created: https://github.com/dequelabs/docs-developer-hub/issues/59~~
<details>
<summary>Example slack notification</summary>

image

</details>

Closes: https://github.com/dequelabs/axe-api-team/issues/386

PR #120, opened 2024-01-09
fix(semantic-pr-footer-v1): allow all github terms

straker:
Closes: https://github.com/dequelabs/axe-api-team-public/issues/107


.github/actions/semantic-pr-footer-v1/src/isValidFooter.ts
```diff
@@ -1,12 +1,26 @@
+const validFooters = [
+  // github terms that link an issue
+  // @see https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword
+  'close: ',
```
Zidious, line 4: It appears GitHub closes the referenced issue with `:` omitted, should we remove the use of `:`? 

example: https://github.com/dequelabs/zidious-testing/pull/125

straker: I don't think so. It's true that it can close without the `:`, but that leads to odd situations where you write a paragraph `"I thought about this to close #40"` and github automatically links the ticket to close. For the footer I'd rather be explicit and put the `:` otherwise you could get away with a footer that is invalid but passes like `"closes the loophole"`

Zidious (approved):
LGTM

PR #131, opened 2024-02-07
fix: add dependency type after dependency

michael-siek:
Using yarn in github action for some reason fails when there is a space prior to the dependency you want to update. 

no qa required


Zidious (approved):
LGTM - tests look like they need to be updated though

Zidious (approved):

PR #133, opened 2024-02-16
fix(.github/actions/create-update-axe-core-pull-request): conditionally run `actions/checkout`

michael-siek:
no qa required


Zidious (approved):
LGTM - I would add some context to why this was needed in your PR description

Issue #140, opened 2024-03-04
Create public GHA for prisma scanning


dequejenn:
Create a new YML file based off of the one used by mobile ( https://github.com/dequelabs/devtools-mobile-service/blob/develop/.github/workflows/checkov.yml) that will send scans to prisma. 

- Please  move the value of `PRISMA_API_URL: https://api4.prismacloud.io` into a secret
- Also make `output_file_path:` a variable passed in by the repo calling the GHA

Issue #115, opened 2024-01-08
tech-debt: Consider moving shared get version from lerna.json or package.json to own action


Zidious:
We use the following snippet for getting a version in more than one action. As it's fairly simple we can move it to its own shell script action: 

```yml
- name: Get version
  shell: bash
  id: get-version
  run: |
    # Get the version number from the prepare_release shell script and set it as an output
    # Monorepos have a lerna.json file that contains the updated version number
    # Non-monorepos, or non-npm repos, have a package.json file that contains the updated version number
    if [ -f "lerna.json" ]; then
      echo "release_candidate_version=$(cat lerna.json | jq -r '.version')" >> $GITHUB_OUTPUT
    else
      echo "release_candidate_version=$(cat package.json | jq -r '.version')" >> $GITHUB_OUTPUT
    fi
```

straker:
If this is it's own action we'll want to make it a node script and add tests. I know it'll be super small and tests super lite, but we shouldn't have any standalone actions not have tests.

PR #141, opened 2024-03-11
ci(.github/action): add checkov-scan github action

maksym-shynkarenko:
Add the GitHub Action checkov-scans-v1.

Closes: #140

### DIFF
--- a/.github/actions/checkov-scans-v1/README.md
+++ b/.github/actions/checkov-scans-v1/README.md
@@ -4,12 +4,11 @@ A GitHub Action to run [Checkov scans](https://github.com/bridgecrewio/checkov-a
 
 ## Inputs
 
-| Name                | Required | Description                       | Default |
-| ------------------- | -------- | --------------------------------- | ------- |
-| `output-file-path`  | Yes      | Output file path of Checkov scans | NA      |
-| `prisma-access-key` | Yes      | Prisma Cloud access key           | NA      |
-| `prisma-secret-key` | Yes      | Prisma Cloud secret key           | NA      |
-| `prisma-api-url`    | Yes      | Prisma Cloud API URL              | NA      |
+| Name                | Required | Description             | Default |
+| ------------------- | -------- | ----------------------- | ------- |
+| `prisma-access-key` | Yes      | Prisma Cloud access key | NA      |
+| `prisma-secret-key` | Yes      | Prisma Cloud secret key | NA      |
+| `prisma-api-url`    | Yes      | Prisma Cloud API URL    | NA      |
 
 ## Example usage
 
@@ -29,7 +28,6 @@ jobs:
     steps:
       - uses: dequelabs/axe-api-team-public/.github/actions/checkov-scans-v1@main
         with:
-          output-file-path: sca/
           prisma-access-key: ${{ secrets.PRISMA_ACCESS_KEY }}
           prisma-secret-key: ${{ secrets.PRISMA_SECRET_KEY }}
           prisma-api-url: ${{ secrets.PRISMA_URL }}
--- a/.github/actions/checkov-scans-v1/README.md
+++ b/.github/actions/checkov-scans-v1/README.md
@@ -1,0 +1,36 @@
+# checkov-scans-v1
+
+A GitHub Action to run [Checkov scans](https://github.com/bridgecrewio/checkov-action/tree/v12#readme) and a result will be shown in Prisma Cloud.
+
+## Inputs
+
+| Name                | Required | Description                       | Default |
+| ------------------- | -------- | --------------------------------- | ------- |
+| `output-file-path`  | Yes      | Output file path of Checkov scans | NA      |
+| `prisma-access-key` | Yes      | Prisma Cloud access key           | NA      |
+| `prisma-secret-key` | Yes      | Prisma Cloud secret key           | NA      |
+| `prisma-api-url`    | Yes      | Prisma Cloud API URL              | NA      |
+
+## Example usage
+
+```yaml
+name: Checkov
+
+on:
+  push:
+    branches:
+      - release
+      - main
+
+jobs:
+  checkov-scan:
+    runs-on: ubuntu-latest
+    timeout-minutes: 2
+    steps:
+      - uses: dequelabs/axe-api-team-public/.github/actions/checkov-scans-v1@main
+        with:
+          output-file-path: sca/
+          prisma-access-key: ${{ secrets.PRISMA_ACCESS_KEY }}
+          prisma-secret-key: ${{ secrets.PRISMA_SECRET_KEY }}
+          prisma-api-url: ${{ secrets.PRISMA_URL }}
+```
--- a/.github/actions/checkov-scans-v1/action.yml
+++ b/.github/actions/checkov-scans-v1/action.yml
@@ -2,9 +2,6 @@ name: checkov-scans
 description: Checkov scans cloud infrastructure configurations to find misconfigurations before they're deployed and sends a result to Prisma Cloud.
 
 inputs:
-  output-file-path:
-    description: Output file path of Checkov scans
-    required: true
   prisma-access-key:
     description: Prisma Cloud access key
     required: true
@@ -30,6 +27,5 @@ runs:
         output_format: cli
         framework: sca_package,github_actions,dockerfile,secrets
         soft_fail_on: LOW
-        output_file_path: ${{ inputs.output-file-path }}
       env:
         PRISMA_API_URL: ${{ inputs.prisma-api-url }}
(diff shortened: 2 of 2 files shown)


.github/actions/checkov-scans-v1/README.md
```diff
@@ -0,0 +1,36 @@
+# checkov-scans-v1
+
+A GitHub Action to run [Checkov scans](https://github.com/bridgecrewio/checkov-action/tree/v12#readme) and a result will be shown in Prisma Cloud.
+
+## Inputs
+
+| Name                | Required | Description                       | Default |
+| ------------------- | -------- | --------------------------------- | ------- |
+| `output-file-path`  | Yes      | Output file path of Checkov scans | NA      |
```
Zidious, line 9: I know Mobile do this, but we shouldn't have to save each scan into some named directory, they are available on Prisma Cloud, seems a bit redundant, thoughts? 

```suggestion
```

.github/actions/checkov-scans-v1/action.yml
```diff
@@ -0,0 +1,40 @@
...
+runs:
+  using: 'composite'
+  steps:
+    # Checks-out your repository under $GITHUB_WORKSPACE, so follow-up steps can access it
+    - uses: actions/checkout@v4
+    # Used to name the file from Checkov
+    - name: Extract package version
+      id: extract-version
+      shell: bash
+      run: echo "::set-output name=version::$(jq -r .version package.json)"
```
Zidious, line 27: This way of setting outputs for GHA's has been deprecated for a while (see [article](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)). There are various examples in the actions directory of apply the new format
```suggestion
			run: echo version=$(jq -r .version package.json) >> "$GITHUB_OUTPUT"
```
The root package JSON doesn't normally update it's version when releasing. We have a script that does get reused quite regularly and there is a tech-debt ticket to turn that into it's own GHA. For now I think it's fine to reuse the script and reference the ticket so we can eventually pay that debt back: https://github.com/dequelabs/axe-api-team-public/issues/115. 

I don't actually see for example: `steps.extract-version.outputs.version` be used in the actual prisma cloud action, do we even need it or can we just run prisma?

.github/actions/checkov-scans-v1/action.yml
```diff
@@ -0,0 +1,40 @@
...
+    - name: Run Prisma Cloud
+      id: prisma-cloud
+      uses: bridgecrewio/checkov-action@v12
+      with:
+        api-key: ${{ inputs.prisma-access-key }}::${{ inputs.prisma-secret-key }}
+        # CLI output to the console and md file
+        output_format: cli
+        framework: sca_package,github_actions,dockerfile,secrets
+        soft_fail_on: LOW
+        output_file_path: ${{ inputs.output-file-path }}
```
Zidious, line 38: See comment about saving these scans - 
```suggestion
```

.github/actions/checkov-scans-v1/action.yml
```diff
@@ -0,0 +1,40 @@
+name: checkov-scans
+description: Checkov scans cloud infrastructure configurations to find misconfigurations before they're deployed and sends a result to Prisma Cloud.
+
+inputs:
+  output-file-path:
+    description: Output file path of Checkov scans
+    required: true
```
Zidious, line 7: See comment about saving these scans -

```suggestion
```

Zidious (changes requested):
Overall LGTM! - just a few minor things inline

.github/actions/checkov-scans-v1/README.md
maksym-shynkarenko: To be honest I don't know, because I couldn't test it in the worked repo. This variable is used for bridgecrewio/checkov-action@v12. Here is from docs. So when I can test it in the worked repo I will remove this param if it's unnecessary.

image

.github/actions/checkov-scans-v1/action.yml
maksym-shynkarenko: This is not used because we don't need a PR after that. I'll delete it

.github/actions/checkov-scans-v1/action.yml
maksym-shynkarenko: commented above

.github/actions/checkov-scans-v1/action.yml
maksym-shynkarenko: @Zidious 
deleted this step

.github/actions/checkov-scans-v1/README.md
Zidious: Ah ok, ya the action will upload the scan results so no need to save them within the repo IMO

.github/actions/checkov-scans-v1/action.yml
```diff
@@ -0,0 +1,35 @@
...
+
+runs:
+  using: 'composite'
+  steps:
+    # Checks-out your repository under $GITHUB_WORKSPACE, so follow-up steps can access it
+    - uses: actions/checkout@v4
+    # The checkov scan
+    - name: Run Prisma Cloud
+      id: prisma-cloud
+      uses: bridgecrewio/checkov-action@v12
```
michael-siek, line 26: We'll need to make sure that we pin this version to a specific commit SHA

michael-siek (changes requested):

.github/actions/checkov-scans-v1/action.yml
maksym-shynkarenko: but we don't need to create any PR or do something with GitHub

straker: It's a [policy for security](https://github.com/dequelabs/axe-api-team/blob/main/policies/github-actions.md#pin-3rd-party-actions-to-commit-sha) to pin 3rd party actions to a commit SHA. Git tags are changeable so `v12` could be updated to include malicious code in the future.

.github/actions/checkov-scans-v1/README.md
maksym-shynkarenko: deleted the input `output-file-path`. Also pushed the changes of workflow for jazzband, axe-linter

.github/actions/checkov-scans-v1/action.yml
Zidious: Yeah good catch - most 3rd party actions should be pinned to a particular hash and have a dependabot identifier so they can get automatically updated like so: 

```
uses: bridgecrewio/checkov-action@99bb2caf247dfd9f03cf984373bc6043d4e32ebf # tag=v12
```

The SHA would would be whatever is latest at the time of creating this PR in this case its 99bb.. on the v12 tag. the `# tag=v12` tells dependabot to update to the latest SHA on the v12 tag if any

Zidious (changes requested):
Just one comment outstanding from Michael and Steve otherwise LGTM!

.github/actions/checkov-scans-v1/action.yml
maksym-shynkarenko: @michael-siek 
set the SHA for the checkov-action

.github/actions/checkov-scans-v1/action.yml
```diff
@@ -0,0 +1,31 @@
...
+
+runs:
+  using: 'composite'
+  steps:
+    # Checks-out your repository under $GITHUB_WORKSPACE, so follow-up steps can access it
+    - uses: actions/checkout@v4
+    # The checkov scan
+    - name: Run Prisma Cloud
+      id: prisma-cloud
+      uses: bridgecrewio/checkov-action@8b268e1e4d7e52a78ba9d86e2489fb09b8a340c3 # tag=v12.2690.0
```
Zidious, line 23: The pinned SHA is not quite accurate you'd want [99bb2caf247dfd9f03cf984373bc6043d4e32ebf](https://github.com/bridgecrewio/checkov-action/commit/99bb2caf247dfd9f03cf984373bc6043d4e32ebf) which is the latest on the [v12](https://github.com/bridgecrewio/checkov-action/releases/tag/v12.1347.0) tag
```suggestion
      uses: bridgecrewio/checkov-action@99bb2caf247dfd9f03cf984373bc6043d4e32ebf # tag=v12
```

Zidious (changes requested):
One very minor thing otherwise this is LGTM

.github/actions/checkov-scans-v1/action.yml
Zidious: Ignore me! This is the latest I missed that they released [yesterday](https://github.com/bridgecrewio/checkov-action/releases/tag/v12.2690.0)

Zidious (approved):
LGTM!

michael-siek (approved):

PR #160, opened 2024-04-24
feat(.github/action): change the label-and-move-released-issues and move-released-issues-and-create-sync-pr actions to change a column name

maksym-shynkarenko:
Added the input `column-name` in the `label-and-move-released-issues` and `move-released-issues-and-create-sync-pr` actions.

Closes: [#505](https://github.com/dequelabs/axe-api-team/issues/505)

### DIFF
--- a/.github/actions/label-and-move-released-issues-v1/src/run.test.ts
+++ b/.github/actions/label-and-move-released-issues-v1/src/run.test.ts
@@ -117,6 +117,7 @@ describe('run', () => {
     version?: string
     token?: string
     projectNumber?: string
+    releaseColumn?: string
   }
 
   const generateInputs = (inputs?: Partial<GenerateInputsArgs>) => {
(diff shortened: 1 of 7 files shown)


.github/actions/label-and-move-released-issues-v1/src/run.test.ts
```diff
@@ -132,12 +133,16 @@ describe('run', () => {
     const projectNumber = getInput
       .withArgs('project-number')
       .returns(inputs?.projectNumber ?? '66')
+    const releaseColumn = getInput
```
Zidious, line 136: Just missing two tests here: 

- When the input `column-name` is supplied ensure we respect that e.g. `some-other-column-name`
- When the inout is not provided ensure we default to `released`

Zidious (changes requested):
One minor thing otherwise LGTM

.github/actions/label-and-move-released-issues-v1/src/run.test.ts
maksym-shynkarenko: as discussed with @Zidious this is not needed

Zidious (approved):

PR #161, opened 2024-04-24
ci(.github/action): change the `create-release-candidate` action

maksym-shynkarenko:
Added to the `create-release-candidate` action the optional `docs-labels` input, changed the project number and column names, and changed the release-issue content.

Closes: [#501](https://github.com/dequelabs/axe-api-team/issues/501)

### DIFF
--- a/.github/actions/create-release-candidate-v1/README.md
+++ b/.github/actions/create-release-candidate-v1/README.md
@@ -12,6 +12,7 @@ A GitHub Action to create a release candidate.
 | `release-script-path` | Yes      | The path to the [release script](#release-script-requirements) that creates the changelogs and bumps the version of the package(s)                                                                                              | NA      |
 | `version-locked`      | No       | Whether or not the version bump should treat major/minor as "locked". Repos which version-lock to axe-core should default this to `true`, overriding it to `false` only for releases that update `axe-core`.                    | `false` |
 | `docs-repo`           | No       | The name of the repo where the release notes live                                                                                                                                                                               | `null`  |
+| `docs-labels`         | No       | Labels for release notes issue. Comma-delimited list of labels (e.x. "release,PRIORITY: high")                                                                                                                                  | NA      |
 | `should-checkout`     | No       | Whether or not the action should checkout the repository.                                                                                                                                                                       | `true`  |
 
 ## Example usage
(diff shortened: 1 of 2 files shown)


.github/actions/create-release-candidate-v1/README.md
```diff
@@ -12,6 +12,7 @@ A GitHub Action to create a release candidate.
 | `release-script-path` | Yes      | The path to the [release script](#release-script-requirements) that creates the changelogs and bumps the version of the package(s)                                                                                              | NA      |
 | `version-locked`      | No       | Whether or not the version bump should treat major/minor as "locked". Repos which version-lock to axe-core should default this to `true`, overriding it to `false` only for releases that update `axe-core`.                    | `false` |
 | `docs-repo`           | No       | The name of the repo where the release notes live                                                                                                                                                                               | `null`  |
+| `docs-labels`         | No       | Labels for release notes issue. Comma-delimited list of labels (e.x. "release,PRIORITY: high")                                                                                                                                  | NA      |
```
Zidious, line 15: When would the labels ever change for an RC issue that's created? I saw in the ticket it's need for Jazzband but that's happening already, example: https://github.com/dequelabs/docs-developer-hub/issues/116. If we want the labels to be repo specific we should add some defaults as I suspect most will not change from having the `release` and `PRIORITY: HIGH` labels?

.github/actions/create-release-candidate-v1/action.yml
```diff
@@ -16,14 +16,18 @@ inputs:
...
     description: 'Whether or not the version should be locked to axe-core'
-    default: 'false'
     required: false
+    default: 'false'
   docs-repo:
     description: 'The name of the repo where the release notes live'
+    required: false
     default: 'null'
+  docs-labels:
+    description: 'Labels for release notes issue. Comma-delimited list of labels (e.x. "release,PRIORITY: high")'
```
Zidious, line 26: Is it expected the default will be no labels moving forward? If not we should add a default value here for `"release,PRIORITY: high"`

Zidious (changes requested):
Had two questions inline -

.github/actions/create-release-candidate-v1/README.md
maksym-shynkarenko: as discussed with @Zidious  this is not needed

.github/actions/create-release-candidate-v1/action.yml
maksym-shynkarenko: as discussed with @Zidious this is not needed

Zidious (approved):
LGTM - 

For future us, discussed offline now we'll have one docs repo that houses all documentation and not all repos that use this flow will want to have `release` and `PRIORITY: high` labels attached by default, the default behaviour will be no labels.

PR #167, opened 2024-05-15
chore: update to support eslint 9 flat config

michael-siek:
Ref: https://github.com/dequelabs/axe-api-team/issues/511
Update to use eslint9 and support the new flat config
Ref: [eslint.org/docs/latest/use/configure/migration-guide](https://eslint.org/docs/latest/use/configure/migration-guide)

no qa required


eslint.config.js
```diff
@@ -0,0 +1,31 @@
...
+
+module.exports = [
+  eslint.configs.recommended,
+  ...tseslint.configs.recommended,
+  {
+    languageOptions: {
+      parser: typescriptParser,
+      globals: {
+        ...globals.node,
+        ...globals.es2015
```
Zidious, line 15: Do we need this `globals.es2015` option?

Zidious (changes requested):
Minor things inline -

eslint.config.js
michael-siek: We had `es6: true` in the previous config file which is the same as `globals.es2015`

Zidious (approved):

PR #172, opened 2024-05-24
ci(.github/action): add release candidate version input in the abort-release-candidate action

maksym-shynkarenko:
Added an optional release candidate version input to abort the exact RC. If this input is not provided the GHA will take the version from `lerna.json` (if it exists) or from the `package.json`.

Closes: [#513](https://github.com/dequelabs/axe-api-team/issues/513)

### DIFF
--- a/.github/actions/abort-release-candidate-v1/README.md
+++ b/.github/actions/abort-release-candidate-v1/README.md
@@ -4,12 +4,12 @@ A GitHub Action to abort a release candidate.
 
 ## Inputs
 
-| Name        | Required | Description                                                                              | Default |
-| ----------- | -------- | ---------------------------------------------------------------------------------------- | ------- |
-| `token`     | Yes      | A GitHub token used for octokit and GH CLI with the [required permissions](#permissions) | NA      |
-| `base`      | Yes      | The base branch the release candidate was going to be merged into                        | NA      |
-| `docs-repo` | No       | The docs repo where the release notes issue is located                                   | NA      |
-| `version`   | No       | The release candidate version number that will be aborted (e.g 1.0.0)                    | NA      |
+| Name        | Required | Description                                                                              | Default                             |
+| ----------- | -------- | ---------------------------------------------------------------------------------------- | ----------------------------------- |
+| `token`     | Yes      | A GitHub token used for octokit and GH CLI with the [required permissions](#permissions) | NA                                  |
+| `base`      | Yes      | The base branch the release candidate was going to be merged into                        | NA                                  |
+| `docs-repo` | No       | The docs repo where the release notes issue is located                                   | NA                                  |
+| `version`   | No       | The release candidate version number that will be aborted (e.g 1.0.0)                    | from `lerna.json` or `package.json` |
 
 ## Permissions
 
--- a/.github/actions/abort-release-candidate-v1/README.md
+++ b/.github/actions/abort-release-candidate-v1/README.md
@@ -4,12 +4,12 @@ A GitHub Action to abort a release candidate.
 
 ## Inputs
 
-| Name        | Required | Description                                                                              | Default                             |
-| ----------- | -------- | ---------------------------------------------------------------------------------------- | ----------------------------------- |
-| `token`     | Yes      | A GitHub token used for octokit and GH CLI with the [required permissions](#permissions) | NA                                  |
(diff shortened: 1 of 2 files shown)


.github/actions/abort-release-candidate-v1/action.yml
```diff
@@ -12,6 +12,9 @@ inputs:
     description: 'The docs repository where the issue for the release candidate is located'
     default: ''
     required: false
+  version:
```
Zidious, line 15: I'm not entirely sure we want this. The abort-release-candidate action should be able to infer the version of the RC it's trying to abort. I think there is a something wrong with _how_ we are not checking out the `release` branch? 

I would imagine if you run `create-release-candidate`, that PR gets merged into `release` and now release holds a bumped version of e.g. `1.1.0`. Checking out `origin/release` and running  this [snippet](https://github.com/dequelabs/axe-api-team-public/pull/172/files#diff-58d5337330f7e542de2ed43b7f979a91c8ea07477f1397b364e19e9816d5ff48L31-L37) should pick up the new bumped version `1.1.0`

Zidious (changes requested):
See comment inline -

.github/actions/abort-release-candidate-v1/action.yml
maksym-shynkarenko: As we discussed with @dequejenn we should be able to abort RC whenever we want.
In the first case, we want to abort RC **before** merging `release-YYYY-MM-DD -> release`. In this case, we provide the exact version because `lerna.json` or `package.json` files have the old version number.
In the second case, we want to abort RC **after** merging into the `release` branch. In this case we will take the version number from `lerna.json` or `package.json` files

Zidious (dismissed):

.github/actions/abort-release-candidate-v1/action.yml
dbjorge, line 15: I agree with @Zidious 's original comment - I don't think this is the right way to solve the problem. If the goal is support for releases that have an unmerged RC PR, why not just look for PRs with `label: release` and infer the version from that?

.github/actions/abort-release-candidate-v1/action.yml
```diff
@@ -28,15 +31,21 @@ runs:
...
+        VERSION="${{ inputs.version }}"
+        
+        if [ -z "$VERSION" ]; then
+          # Get the version number from the prepare_release shell script and set it as an output
+          # Monorepos have a lerna.json file that contains the updated version number
+          # Non-monorepos, or non-npm repos, have a package.json file that contains the updated version number
+          if [ -f "lerna.json" ]; then
+            VERSION=$(cat lerna.json | jq -r '.version')
+          else
+           VERSION=$(cat package.json | jq -r '.version')
```
dbjorge, line 43:
```suggestion
            VERSION=$(cat package.json | jq -r '.version')
```

.github/actions/abort-release-candidate-v1/action.yml
```diff
@@ -28,15 +31,21 @@ runs:
...
+          fi
         fi
 
+        echo "release_candidate_version=$VERSION" >> $GITHUB_OUTPUT
```
dbjorge, line 47: This should also echo to stdout the value it found for debugging purposes

dbjorge (changes requested):

.github/actions/abort-release-candidate-v1/action.yml
dbjorge: (could maybe generalize `is-release-in-progress` to `get-release-in-progress` to be able to reuse it)

.github/actions/abort-release-candidate-v1/README.md
```diff
@@ -9,6 +9,7 @@ A GitHub Action to abort a release candidate.
 | `token`     | Yes      | A GitHub token used for octokit and GH CLI with the [required permissions](#permissions) | NA      |
 | `base`      | Yes      | The base branch the release candidate was going to be merged into                        | NA      |
 | `docs-repo` | No       | The docs repo where the release notes issue is located                                   | NA      |
+| `version`   | No       | The release candidate version number that will be aborted (e.g 1.0.0)                    | NA      |
```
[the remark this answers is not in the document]
maksym-shynkarenko: This action just closes RC issue and RC docs issue

dbjorge (approved):

PR #187, opened 2024-07-10
fix: update default board to point to new bass team

Zidious:
Blame jenn - 🍐 _forced_ me to update to 186 (bass team board).

No QA Required


stephenmathieson (approved):
🤷

PR #199, opened 2024-10-16
chore(actions): Update board to latest

dequejenn:
No qa required


maksym-shynkarenko (changes requested):
The project 186 does not exist and the new one (188) has been changed and merged in this workflow file in the [PR](https://github.com/dequelabs/axe-api-team-public/pull/213). So we don't need these changes anymore

PR #203, opened 2024-10-30
fix: do not move issue to column if none is provided

### LINKED ISSUE #200: Remove forced move to column
In .github/actions/label-and-move-released-issues-v1 please make it so if a column is supplied in the called action, the issue moves to the column supplied. If no column is supplied, then the issue does not move (instead of current behavior of having a default culmn)

scottmries:
Closes #200


.github/actions/label-and-move-released-issues-v1/action.yml
```diff
@@ -17,7 +17,7 @@ inputs:
   column-name:
     description: 'Name of column to move to'
     required: false
-    default: 'released'
```
dbjorge, line 20: Can you update this in the `README` as well and clarify there that leaving it unspecified will result in skipping the column movement?

dbjorge (dismissed):

dbjorge (approved):

PR #204, opened 2024-12-20
ci: add require multiple reviewers action

zlayaAvocado:
Requires a specified number of reviewers for certain important files, the files are to be provided as an `important-files-path` var. 

See `README.md` for more details https://github.com/dequelabs/axe-api-team-public/pull/204/files#diff-db32ad4c1ae8b5f09def402ee7d7876046a269ae8c7b5bfb42567d703807902b

**NOTE:** this version doesn't account for edge cases like accounting for approvals from people who requested changes, e.g. if a PR requires 2 reviewers, 1 reviewer requested changes but other 2 reviewers approved the PR, the action will pass.

1. This would show a `Tests / require-multiple-reviewers` check that would fail until a required number of reviewers approves a PR that uses the action.
2. Also shows annotations on the files that are important. Annotations remain on the files for information purposes and are not removed.

**Changes requested:**
Tests / require-multiple-reviewers` fails since the PR requires 1 reviewer and changes are requested

**Changes approved:**
Tests / require-multiple-reviewers` passes since the PR requires 1 reviewer and it's approved by 1 reviewer

**Annotations:**
`Annotation on an important file 'The file important.txt is important and requires at least 1 reviewers.'

Ref: https://github.com/dequelabs/axe-linter/issues/1632


.github/actions/require-multiple-reviewers-v1/src/run.ts
```diff
@@ -0,0 +1,63 @@
...
+    const octokit = github.getOctokit(token)
+    const { owner, repo } = github.context.repo
+    const requiresMultipleReviewers = importantFilesChanged.length > 0
+    let conclusion: Conclusion = 'neutral'
+
+    if (requiresMultipleReviewers) {
+      const { data: reviews } = await octokit.rest.pulls.listReviews({
+        owner,
+        repo,
+        pull_number: github.context.payload.pull_request!.number
```
dbjorge, line 39: If someone tries to add this to a non-pull_request-triggered-workflow (eg, they try adding it in a step of an existing `push`-triggered CI workflow or something), it'd be good to include a more specific error message than the "tried to access property `number` of undefined" I imagine this will emit

.github/actions/require-multiple-reviewers-v1/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+A GitHub Action that requires multiple reviewers for important files
+
+## Inputs
+
+| Name                   | Required | Description                                                | Default |
+| ---------------------- | -------- | ---------------------------------------------------------- | ------- |
+| `token`                | Yes      | A GitHub token with the required permissions               | NA      |
+| `number-of-reviewers`  | Yes      | The number of reviewers required.                          | 2       |
+| `changed-files-path`   | Yes      | The path to the file containing a list of changed files.   | NA      |
```
dbjorge, line 11: Can this be more specific about the expected formats of the files? How is the list separated? What are the file paths relative to? Does it support any sort of globbing/patterns? (it's ok for this to start simple but it should be documented)

dbjorge (changes requested):

.github/actions/require-multiple-reviewers-v1/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Get changed files for the entire PR
+        run: |
+          # Get the changed files by comparing base and head branches
+          git diff --name-only origin/${{ github.event.pull_request.base.ref }}...origin/${{ github.event.pull_request.head.ref }} > .github/changed-files.txt
+
+      - name: Require two reviewers for important files
+        uses: dequelabs/axe-api-team-public/.github/actions/require-multiple-reviewers-v1@main
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
zlayaAvocado: The secrets.GITHUB_TOKEN isn’t available to actions, it must be passed explicitly: https://github.com/actions/toolkit/blob/main/packages/github/README.md?plain=1#L14-L18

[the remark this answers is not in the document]
dbjorge: It's not available implicitly via `secrets.GITHUB_TOKEN`, but their docs suggest that it is available implicitly via the `github.token` context. I just verified in a test repo that this does appear to be the case: https://github.com/dequelabs/dbjorge-test-repo/actions/runs/12469818616/job/34803729873

.github/actions/require-multiple-reviewers-v1/README.md
```diff
@@ -21,28 +20,15 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 2
     permissions:
-      pull-requests: write
+      pull-requests: read
       contents: read
+      checks: write
     steps:
```
dbjorge, line 26: A checkout step of some sort is still required in order to read `important-files.txt`, no?

.github/actions/require-multiple-reviewers-v1/src/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+  IMPORTANT_FILES_PATH: string,
+  changedFiles: string[]
+): string[] {
+  // Since this is the ignore package, it will filter out files that match the patterns
+  const notImportantFiles = ignore()
+    .add(fs.readFileSync(IMPORTANT_FILES_PATH, 'utf-8').toString())
+    .filter(changedFiles)
+
+  // Get all the files that were filtered out earlier - those are the important files
+  return changedFiles.filter(file => !notImportantFiles.includes(file))
```
dbjorge, line 39: This is `O(unimportant_file_count * total_changed_file_count)`, which might be very large for a PR that touches lots of unimportant files. Can you convert `notImportantFiles` to a `Set` in between L36 and L39 to avoid the quadratic worst-case?

package.json
```diff
@@ -40,6 +40,7 @@
     "glob": "^10.4.2",
     "globals": "^15.7.0",
     "husky": "^9.0.11",
+    "ignore": "^6.0.2",
```
dbjorge, line 43: :+1:, verified license and maintenance record

dbjorge (changes requested):

.github/actions/require-multiple-reviewers-v1/src/run.ts
```diff
@@ -0,0 +1,72 @@
...
+      })
+
+      const approvals = reviews.filter(review => review.state === 'APPROVED')
+      conclusion = approvals.length < numberOfReviewers ? 'failure' : 'success'
+    }
+
+    await octokit.request('POST /repos/{owner}/{repo}/check-runs', {
```
dbjorge, line 55: Did you test that this does the right thing in the event that we already left a check for the same commit? eg, if it's triggered from a `pull_request` and then later a `pull_request_review` trigger, did you verify that the second one looks like it's updating the original check status (as opposed to leaving a new second status)? What about the annotations - do they get cleared out when posting the updated check, or do they need to be removed explicitly?

(I'd recommend trying this out in a separate test repo and linking the results of your testing in the PR description)

dbjorge (dismissed):

.github/actions/require-multiple-reviewers-v1/src/run.ts
zlayaAvocado: I'ts going to fail and re-run the `Tests / require-multiple-reviewers` action 

image

Annotations are warnings and don't go anywhere, they are for the information purpose so a PR author knows what important files are, and I don't think they should be removed.

.github/actions/require-multiple-reviewers-v1/src/utils.ts
zlayaAvocado: Good point, done some reading about it too :D I updated the logic to do less filtering, let me know if any array still should be a `Set` but I am not sure how it would help in this case.

stephenmathieson:
Does `.github/CODEOWNERS` not already do this? 🤔

zlayaAvocado:
> Does `.github/CODEOWNERS` not already do this? 🤔

I think with codeowners you need to request specific people if you want a PR to be reviewed and approved more than once, something like `*.py  @user1 @user2` where these users will be automatically assigned to any `.py` file changed. This PR requests any specified number of reviewers to approve the files you specify, and it's not tied to any particular teams or people.

That is to my understanding though, let me know if I am wrong 😄

dbjorge:
> Does .github/CODEOWNERS not already do this? 🤔

> That is to my understanding of codeowners though, let me know if I am wrong 😄

I agree with your interpretation. GitHub's implementation of CODEOWNERs unfortunately doesn't support requiring multiple reviewers; you can specify a file as being owned by multiple entities, but that results in GitHub requiring a review from any one of those entities (not each of those entities). From [GitHub's docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-and-branch-protection):

> Note
>
> When reviews from code owners are required, an approval from any of the owners is sufficient to meet this requirement. For example, let's say that your CODEOWNERS file contains the following line:
>
> `*.js     @global-owner1 @global-owner2`
> This means that changes to JavaScript files could be approved by either `@global-owner1` or `@global-owner2`, but approvals from both are not required.

CODEOWNERS isn't a particularly well-defined format, and there exist other managed Git platforms that support more advanced CODEOWNERS features (including this). For example, [GitLab supports this as part of their CODEOWNERS section syntax](https://docs.gitlab.com/ee/user/project/codeowners/#require-multiple-approvals-from-code-owners). Unfortunately, support for this sort of thing [doesn't seem to have gained any traction at GitHub](https://github.com/orgs/community/discussions/61357).

I think using special comments in CODEOWNERS or something instead of a new file would be a fine interface for this to use, but I didn't suggest it originally because I worried I'd be getting into bikeshedding territory.

.github/actions/require-multiple-reviewers-v1/src/run.ts
dbjorge: I'm fine with deferring this and leaving it as-is for the purposes of this PR. I suspect this is going to feel noisy, but we can adjust that once we see it in action if we need to.

.github/actions/require-multiple-reviewers-v1/src/utils.test.ts
```diff
@@ -177,4 +181,101 @@ describe('utils', () => {
...
+          submitted_at: '2023-01-03T00:00:00Z'
+        }
+      ]
+
+      const result = getApproversCount(reviews as Array<Review>)
+
+      expect(result).to.equal(2)
+    })
+
+    it('should handle multiple reviews from the same user and count only the latest approval', () => {
```
dbjorge, line 210: Good catch

dbjorge (approved):

PR #213, opened 2025-04-07
chore(.github/action): change `project-number`, `column-name`, Slack channel for the new team-results team

maksym-shynkarenko:
Changed all the action and workflow files to work with the new `team-results` board and Slack channel

Closes: [#4](https://github.com/dequelabs/results-team/issues/4)

### DIFF
--- a/.github/actions/abort-release-candidate-v1/action.yml
+++ b/.github/actions/abort-release-candidate-v1/action.yml
@@ -65,13 +65,13 @@ runs:
       env:
         GH_TOKEN: ${{ inputs.token }}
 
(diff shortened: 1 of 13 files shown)


scurker (commented):
It feels weird having this live in `axe-api-team-public` which is now defunct. Have we discussed renaming this repo to reflect the new team's ownership?

maksym-shynkarenko:
> It feels weird having this live in `axe-api-team-public` which is now defunct. Have we discussed renaming this repo to reflect the new team's ownership?

This repo exists and is used by other repos. We can rename it but after that, we will have to change the source in every action and workflow file

scurker (approved):

PR #214, opened 2025-04-17
fix(checkov-scans-v1): allow frameworks as input

### LINKED ISSUE #212: Amend the checkov action to take optional input to override frameworks which are scanned
At the moment we have [hardcoded the frameworks](https://github.com/dequelabs/axe-api-team-public/blob/main/.github/actions/checkov-scans-v1/action.yml#L28) which are scanned by Checkov: 

```
framework: sca_package,github_actions,dockerfile,secrets
```

When used in repositories that do not have a `Dockerfile` causes the action to fail vs ignoring the framework. As a workaround we should implement a optional input that can override the default list of frameworks for repos. an Example: 

```
framework_overrides: string
default: sca_package,github_actions,dockerfile,secrets
```

Impl: 

```
// When called within a workflow would override the defaults
framework_overrides: sca_package,github_actions,secrets
```

straker:
Closes #212

### DIFF
--- a/.github/actions/checkov-scans-v1/action.yml
+++ b/.github/actions/checkov-scans-v1/action.yml
@@ -11,6 +11,10 @@ inputs:
   prisma-api-url:
     description: Prisma Cloud API URL
     required: true
+  frameworks:
+    description: Comma-separated list of frameworks to scan
+    required: false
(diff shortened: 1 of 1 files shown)


Zidious (approved):
LGTM

PR #216, opened 2025-04-26
refactor(action): add inputs to the GHA `check-and-move-issue-based-on-labels-v1`

maksym-shynkarenko:
Refactored the GHA `check-and-move-issue-based-on-labels-v1` by adding inputs.

No QA needed.

### DIFF
--- a/.github/actions/check-and-move-issue-based-on-labels-v1/action.yml
+++ b/.github/actions/check-and-move-issue-based-on-labels-v1/action.yml
@@ -13,6 +13,12 @@ inputs:
   issue-number:
     description: 'The issue number to check and move if it matches any'
     required: true
+  issue-owner:
+    description: 'The issue organization name'
+    required: true
+  issue-repo:
+    description: 'The issue repository name'
+    required: true
   team-label:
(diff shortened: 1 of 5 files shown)


.github/actions/check-and-move-issue-based-on-labels-v1/action.yml
```diff
@@ -13,6 +13,12 @@ inputs:
   issue-number:
     description: 'The issue number to check and move if it matches any'
     required: true
+  issue-owner:
+    description: 'The issue organization name'
```
Bracciata, line 17: I question whether these should be `'The issue's organization name'`. I think it is fine as is, though.

Bracciata (approved):
I am also curious how we can test this one without merging if it is possible!

.github/actions/check-and-move-issue-based-on-labels-v1/action.yml
maksym-shynkarenko: Agree because an issue has an owner (author)

Zidious (approved):

Issue #39, opened 2023-10-04
Add update-axe-core 


straker:
We already have https://github.com/dequelabs/update-axe-core, but we should port it to this repo and update it to use TypeScript, add tests, and resolve all the open issues. We should also create a composite action that uses it and opens a pr for us as updating axe-core without a pr is kinda useless. Plus every repo that uses the action has to duplicate the pr creation part.

straker:
Duplicate of #30

PR #219, opened 2025-05-01
fix(action): handle the Octokit label creation if a label exists in the label-released-issues-v1 GHA

maksym-shynkarenko:
Handled the Octokit label creation if a label exists in the label-released-issues-v1 GHA. In some cases, Octokit doesn't see the created label in the same repo in the [previous iteration](https://github.com/dequelabs/jazzband/actions/runs/14763302563/job/41449294781#step:6:56) and throws an error on the [second iteration](https://github.com/dequelabs/jazzband/actions/runs/14763302563/job/41449294781#step:6:172), so we should handle this error.

Closes: [results-team: #39](https://github.com/dequelabs/results-team/issues/39)

### DIFF
--- a/.github/actions/label-released-issues-v1/src/run.ts
+++ b/.github/actions/label-released-issues-v1/src/run.ts
@@ -1,3 +1,4 @@
+import { RequestError as OctokitError } from '@octokit/request-error'
 import type { Core, GitHub } from './types'
 import { ParsedCommitList } from '../../generate-commit-list-v1/src/types'
 import getIssueProjectInfo from '../../label-and-move-released-issues-v1/src/getIssueProjectInfo'
@@ -7,6 +8,11 @@ interface IssueData {
   number: number
   owner: string
   repo: string
+}
+interface OctokitDataError {
+  resource: string
+  code: string
+  field: string
 }
 
 const DEFAULT_DONE_COLUMNS = 'done,devDone'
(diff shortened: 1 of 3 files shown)


.github/actions/label-released-issues-v1/src/run.ts
```diff
@@ -150,12 +156,32 @@ export default async function run(core: Core, github: GitHub): Promise<void> {
...
+
+              if (alreadyExistsError) {
+                shouldThrowError = false
+              }
+            }
+
+            if (shouldThrowError) {
+              throw err
+            }
+          }
```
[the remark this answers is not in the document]
maksym-shynkarenko: we need to check the exact error code `already_exists` from octokit. Also, we don't need the early-return because we should process the error if it's not the error with the code `already_exists`

.github/actions/label-released-issues-v1/src/run.ts
```diff
@@ -150,12 +156,32 @@ export default async function run(core: Core, github: GitHub): Promise<void> {
             `\nThe label "${LABEL}" does not exist for the issue repo ${issueOwner}/${issueRepo}, creating...`
           )
 
-          await octokit.rest.issues.createLabel({
-            repo: issueRepo,
-            owner: issueOwner,
-            name: LABEL,
-            color: 'FFFFFF' // "white" color
-          })
+          // Should be in try-catch to avoid failing the action if the label already exists
```
Zidious, line 159: This is not accurate - the [API](https://octokit.github.io/rest.js/v21/)is paginated capped at around 100 labels / page. The referenced repo where this failed has 149 labels, I suspect if we paginate the request [here ](https://github.com/dequelabs/axe-api-team-public/blob/61e6d34bc88e6309c9a7f9ea68403da6299a89d7/.github/actions/label-released-issues-v1/src/run.ts#L147-L150)in order to locate the label

Zidious (changes requested):

.github/actions/label-released-issues-v1/src/run.ts
maksym-shynkarenko: yes, my bad, added the pagination

Zidious (dismissed):

PR #225, opened 2025-08-19
chore: Update CODEOWNERS

straker:
Updating with shared ownership and changing to the `ocarina-team` rather than the `api-team`

No QA required


Garbee (approved):
Per discussion on slack, this is valid.

Ref: https://deque.slack.com/archives/C05F53ZHHPG/p1755554109276439

Bracciata (approved):

PR #248, opened 2025-10-03
chore: update generated files

straker:
Looks like almost every directory had dist changes that haven't been being generated. Not sure why that is as the `lint-staged` hook should update it before the pr is created and the `sync-generated-files` action should catch any stragglers. 

Could be that updating dependencies could cause changes but would not regenerate dist of our actions if the source files didn't change.

No QA required


Garbee (approved):
Seems impossible to actually review.

I think once we have these running on Node 22 where type stripping is possible we can stop tracking of build files entirely since we could just run the scripts directly.

PR #267, opened 2025-10-03
chore: revert schedule, add comment

straker:
This seems to be working so reverting the `schedule` to monthly and add a comment to why we needed to specify the globstar

No QA needed


straker:
Lets merge this _after_ all the dependabot updates for the `actions` get merged just in case dependabot does something weird after the schedule change.

Issue #238, opened 2025-09-08
Provide strict mode option for trailer syntax


Garbee:
Currently some footers like `qa notes` are allowed in, even though they go against conventional commit spec (rule 9 in this case.)

In strict mode, spaces in keys except for `BREAKING CHANGE` must not be allowed. The extra things we do allow in the footer currently, should be required to be above it in the body of the description before trailers begin.

Issue #237, opened 2025-09-08
Add issue reference strict mode


Garbee:
Some of our tooling seems to bail on commits with links as the `Fixes: ` trailer. Apply a strict mode to look for [github's syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) when these are present.

Return a non-0 status with a useful error (and ideally, parse a correct form to use to provide back to a PR submitter.)

PR #301, opened 2026-01-06
fix(semantic-pr-footer): allow whitespace before footer

straker:
A recent pr into axe-core-npm was failing the semantic footer action. After a bit of confusion I realized it was because there was some whitespace before the `Closes:` text. I don't think we should care about whitespace so am trimming it before processing.

No QA required

### DIFF
--- a/.github/actions/semantic-pr-footer-v1/dist/index.js
+++ b/.github/actions/semantic-pr-footer-v1/dist/index.js
@@ -22,68 +22,36 @@ var __setModuleDefault = (this && this.__setModuleDefault) || (Object.create ? (
 }) : function(o, v) {
     o["default"] = v;
 });
-var __importStar = (this && this.__importStar) || (function () {
-    var ownKeys = function(o) {
-        ownKeys = Object.getOwnPropertyNames || function (o) {
-            var ar = [];
-            for (var k in o) if (Object.prototype.hasOwnProperty.call(o, k)) ar[ar.length] = k;
-            return ar;
-        };
-        return ownKeys(o);
-    };
-    return function (mod) {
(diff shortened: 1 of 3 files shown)


maksym-shynkarenko (approved):

Garbee (changes requested):
Whitespace *is* important. The point of conventional commits is to adhere to the [git trailer](https://git-scm.com/docs/git-interpret-trailers.html) syntax. A space at the start explicitly changes the interpretation.

Working:

```shell
echo "Subject

Reviewed-by: Alice <alice@example.com>
Tested-by: Bob <bob@example.com>" | git interpret-trailers --parse
```

Will output:

```shell
Reviewed-by: Alice <alice@example.com>
Tested-by: Bob <bob@example.com>
```

Not working:

```shell
echo "Subject

Reviewed-by: Alice <alice@example.com>
  Tested-by: Bob <bob@example.com>" | git interpret-trailers --parse
```

Outputs:

```shell
Reviewed-by: Alice <alice@example.com> Tested-by: Bob <bob@example.com>
```

# Alternative

Instead of trimming to do the work at all, we could trim and compare. If the string after `trimStart` is different than before, log that. Or if the token has permission to write to PRs, even comment back on the PR that the description has invalid syntax for the footer.

Garbee:
@straker We do care. Because the scripts can be improved later to use the trailer parsing to find the information we need more easily. (Which has already been brought up before as a way to improve these, which #237 and #238 exist to try and work on as time permits.) As this starts to come into play with the Zendesk integration support being more readily enforced.

If we say now, "Well, we don't care about spaces", then we are stuck custom-coding parsing instead of just relying on the git primitive to do it.

PR #404, opened 2026-06-22
build: upgrade to TypeScript 6 and finalize ESM migration

### LINKED ISSUE #361: Convert to ESM
Migrate to ESM for the builds so we can update deps and not have problems later.

Garbee:
## Overview

Final config-only flip for the ESM + `node:test` migration (#361). Now that all
20 buildable action workspaces have been migrated individually, this PR flips
the root over to ESM, upgrades TypeScript to v6, and removes the legacy test
tooling that the migration made obsolete.

## What changed

**Root → ESM**
- `package.json`: `"type": "module"` + `"engines": { "node": ">=24" }`.
- `tsconfig.json`: `module: esnext` / `moduleResolution: bundler` / `noEmit`,
  with an explicit `"types": ["node"]`. The node types were previously only
  resolving transitively through `@actions/github`, so workspaces that don't
  import it (e.g. `get-new-env-vars-v1`) failed typecheck without this.
- The 20 per-package `tsconfig.json` overrides are reduced to `extends` +
  `include`/`exclude`; their interim ESM compiler options now come from root.

**TypeScript 5.9 → 6.**

**Tooling removal** (replaced by `node:test` + `tsx` + esbuild in the
per-package phase): `mocha`, `nyc`, `chai`, `sinon`, `proxyquire`, `ts-node`,
`@vercel/ncc`, `esbuild-plugin-license` (unused — `build-action.mjs` emits the
license summary itself), and the `@types` for the removed libs. `.mocharc.json`
deleted.

**ESM-only dependency bumps** (these were pinned back in `dependabot.yml`):
`@actions/core` 2→3, `@actions/exec` 2→3, `@actions/github` 8→9,
`conventional-commits-parser` 5→6. `@types/conventional-commits-parser` dropped
(cc-parser 6 self-types). The dependabot "prevent ESM-only" ignore block is
removed.

**Source changes forced by the ESM-only majors**
- `@actions/core` / `@actions/github` no longer have a default export, so
  default imports become namespace imports (`import * as core`).
- `conventional-commits-parser` removed `sync()`; `getCommitType` now uses
  `new CommitParser(options).parse(title)`.

**dist** bundles rebuilt against the updated dependencies.

## Verification

Local, node 24:
- `npm run lint -ws` — clean
- `npm run typecheck` — clean across all 20 workspaces (TS 6)
- `npm run test` — 385/385 passing, coverage thresholds met
- `npm run build` — all 20 workspaces bundle cleanly

Closes: #361

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,10 +1,12 @@
 {
   "compilerOptions": {
     "target": "es2024",
-    "module": "commonjs",
+    "module": "esnext",
+    "moduleResolution": "bundler",
+    "allowImportingTsExtensions": true,
+    "noEmit": true,
+    "types": ["node"],
     "lib": ["dom", "es2024"],
-    "declaration": true,
-    "outDir": "dist",
     "noImplicitReturns": true,
     "esModuleInterop": true,
     "noFallthroughCasesInSwitch": true,
@@ -12,10 +14,8 @@
     "forceConsistentCasingInFileNames": true,
     "pretty": true,
     "strict": true,
-    "sourceMap": true,
     "noImplicitAny": true,
-    "resolveJsonModule": true,
-    "removeComments": true
+    "resolveJsonModule": true
   },
   "exclude": ["**/*.test.ts", "**/test/**/*.ts"]
 }
--- a/.github/actions/add-to-board-v1/dist/index.js
+++ b/.github/actions/add-to-board-v1/dist/index.js
@@ -12,9 +12,6 @@ var __require = /* @__PURE__ */ ((x) => typeof require !== "undefined" ? require
   if (typeof require !== "undefined") return require.apply(this, arguments);
   throw Error('Dynamic require of "' + x + '" is not supported');
 });
-var __esm = (fn, res) => function __init() {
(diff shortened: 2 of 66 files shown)


tsconfig.json
```diff
@@ -1,21 +1,21 @@
...
     "esModuleInterop": true,
     "noFallthroughCasesInSwitch": true,
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
     "pretty": true,
     "strict": true,
-    "sourceMap": true,
     "noImplicitAny": true,
-    "resolveJsonModule": true,
-    "removeComments": true
```
Zidious, line 18: NBD - Did you mean to remove this option?

Zidious (approved):

tsconfig.json
Garbee: Yea, since this does not emit anything it is dead anyways. `ncc` and now `esbuild` are what do all that work. Typescript does not bundle.

chutchins25 (approved):
Talked with Garbee about the git ignore including .nyc_output files after this PR. Looks good to me.

PR #410, opened 2026-06-30
docs: update README.md file

### LINKED ISSUE #273: Make README file more usable
## Description
Review the readme file and make it more usable for newcomers (estimation 1 ST). Add repo description and architecture, approach about how to create new GHAs using `yaml` or `ts` files, and how to test before creating a PR.

maksym-shynkarenko:
This pull request significantly expands and clarifies the `README.md` to provide comprehensive documentation for the repository. The updated README now details the repository's purpose, structure, action development guidelines, and best practices for building, testing, and logging in GitHub Actions. The documentation also includes concrete examples and decision guides for creating both composite (YAML) and Node (TypeScript) actions.

Key documentation improvements:

**Repository structure and usage:**
* Clearly describes the repository as a central library of reusable GitHub Actions, explains how other repositories consume these actions, and outlines the directory structure and use of npm workspaces.

**Action development guidelines:**
* Provides detailed instructions for creating new actions, including naming conventions, directory layout, and versioning strategy (using new directories for breaking changes).
* Explains the difference between composite and Node actions, with a decision table and multiple real-world examples for each.

**Best practices and patterns:**
* Outlines a clear pattern for structuring Node action source code, including validation, error handling, and dependency injection, with code samples.
* Emphasizes required logging practices, security considerations (not logging secrets), and the importance of unit tests with 100% coverage for Node actions.

No QA needed.

Closes: [axe-api-team-public#273](https://github.com/dequelabs/axe-api-team-public/issues/273)

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,8 @@ runs:
 
 Other examples: [cancel-if-release-pr-exists](.github/actions/cancel-if-release-pr-exists/action.yml), [create-release-candidate-v1](.github/actions/create-release-candidate-v1/action.yml) (long but still pure orchestration of other actions).
 
+See GitHub's [composite action docs](https://docs.github.com/actions/creating-actions/creating-a-composite-action) for the full syntax.
+
 ### 2. Node action (TypeScript)
 
 A TypeScript project compiled to a single bundled `dist/index.js` that the runner executes via
@@ -95,17 +97,32 @@ parsing), [has-auto-releasable-commits-v1](.github/actions/has-auto-releasable-c
 
 ### Choosing between them
 
-| If the action mainly…                                        | Use           |
-| ------------------------------------------------------------ | ------------- |
-| calls other actions / `gh` / `git` and passes values through | **Composite** |
-| runs a few shell lines with no parsing                       | **Composite** |
-| queries the GitHub API with custom GraphQL/REST              | **Node (TS)** |
-| parses, transforms, or validates structured data             | **Node (TS)** |
-| has conditionals/policy logic you'd want to unit-test        | **Node (TS)** |
+There is no single rule that always gives the right answer — it is a trade-off. Prefer the simplest
+approach that fits the situation, and weigh the following.
 
-Rule of thumb: **YAML for wiring, TypeScript for logic.** When a composite action's `run:` blocks
-start growing `if`/loops/parsing, that logic belongs in a Node action (often a small one the
-composite then calls).
+**A composite (YAML) action** runs in the **caller's job context** — its tools, Node version, and
+environment are the caller's. That makes it debuggable and fixable on the spot, and steps stay
+visible and overridable without rebuilding a bundle. It fits orchestration and shell work that
+`gh`, `git`, `jq`, or `yq` already do well.
+
+**A Node (TypeScript) action** runs in its **own isolated process with its own bundled
+dependencies**, so it behaves the same everywhere and is highly reproducible. Our established test
+infrastructure (Node test runner, 100% coverage) is TypeScript-only today, so complex logic that
+needs unit tests is easier to maintain here.
+
+**Isolation cuts both ways.** That same isolation removes the caller's ability to adjust behavior
+when something breaks. For example, [update-axe-core-v1](.github/actions/update-axe-core-v1)
+installs dependencies, and in some repos that triggered Playwright to download browsers, hitting a
+Node bug that stalled archive extraction — with **no escape hatch** for the caller. The remedy is a
+composite variant that runs in the caller's Node context and hands them control. During a wide,
+multi-repo release push, "fixable on the spot" can matter more than "perfectly idempotent."
+
+No single signal decides it: `gh` covers most GitHub API calls from a shell, `jq`/`yq` cover
+JSON/YAML, and behavior can be tested in any language — so "uses the API", "parses structured
+data", or "needs tests" are not on their own reasons to reach for Node. Size isn't the deciding
+factor either: a large script should be broken into smaller, focused scripts, just as a large Node
+action is split into small modules. Choose the approach — and the abstraction layer — that makes
+the action easiest to run and maintain **in the environments that call it**.
 
 ## Creating a new action
 
@@ -147,8 +164,9 @@ outputs:
 
 ### The Node action source pattern
 
-Keep `index.ts` a one-liner that injects dependencies into `run`, so `run` and the helpers stay
-pure and testable (no hidden globals):
+Keep `index.ts` a thin entry point that wires the real dependencies and calls `run()`. Because that
+call runs at module load, the logic lives in `run.ts` so a test can `import` it without executing
+the action:
 
 ```typescript
 // src/index.ts
@@ -204,10 +222,16 @@ export default async function run(core: Core, github: Github): Promise<void> {
 
 Split each distinct step into its own `<feature>.ts` with a colocated `<feature>.test.ts` (see
 [generate-commit-list-v1/src](.github/actions/generate-commit-list-v1/src) for the example).
-Tests must hit **100% coverage** of
-`src` (lines, branches, functions) — `index.ts` and `*.test.ts` are excluded.
+Tests must hit **100% coverage** of `src` (lines, branches, functions). Coverage thresholds and
+exclusions are set per action in its `node.config.json` (commonly `index.ts` and `*.test.ts`, and
+sometimes `types.ts`).
 
-### Logging (required)
+Tests use Node's built-in runner with `mock.module` to stub imports (`@actions/exec`, the GitHub
+client, …), then `await import('./run.ts')` and call `run` with lightweight `core`/`github` stubs —
+see
+[generate-commit-list-v1/src/run.test.ts](.github/actions/generate-commit-list-v1/src/run.test.ts).
+
+#### Logging (required)
 
 Every Node action **must** log its progress with `core.info` so that a failed run shows _which
 stage_ it reached, _where_ it stopped, and _why_. Add a log line before each meaningful step (input
@@ -230,6 +254,10 @@ instead. GitHub automatically masks values that came from `${{ secrets.* }}`, sh
 in the run log, but that masking is a safety net, not permission: keep secrets and tokens out of
 your `core.info` / `core.setFailed` strings entirely.
 
+For values that are not `${{ secrets.* }}` but still should not leak (e.g. auto-generated tokens),
+call `core.setSecret(value)` to register them for masking (`::add-mask::`) so the runner redacts
+them from logs.
+
 ### Build (Node actions only)
 
 `npm run build` runs the shared [`scripts/build-action.mjs`](scripts/build-action.mjs), which uses
@@ -246,6 +274,7 @@ Run these from the action's directory (or use `--workspace=<action-name>` from t
 
 ```bash
 npm run lint        # eslint
+npm run lint -ws    # from the repo root (no aggregate `lint` script there)
 npm run typecheck   # tsc --noEmit
 npm test            # Node's built-in test runner; enforces 100% coverage
 npm run build       # rebuild dist/ — then commit it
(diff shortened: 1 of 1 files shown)


straker (dismissed):

README.md
```diff
@@ -1,17 +1,266 @@
...
+  main: 'dist/index.js'
+```
+
+Examples: [move-issues-based-on-label-v1](.github/actions/move-issues-based-on-label-v1) (GraphQL
+project queries), [add-to-board-v1](.github/actions/add-to-board-v1) (wraps the `gh` CLI and
+parses its JSON), [generate-commit-list-v1](.github/actions/generate-commit-list-v1) (commit
+parsing), [has-auto-releasable-commits-v1](.github/actions/has-auto-releasable-commits-v1)
+(release-policy logic).
+
+### Choosing between them
```
Garbee, line 98: One problem in this section, is things get far more nuanced. There is no genuinely clear rule of thumb. For example:

We have the [update-axe-core-v1](https://github.com/dequelabs/axe-api-team-public/tree/main/.github/actions/update-axe-core-v1) action. This has been fine enough, until recently. We hit a problem in some repos where this could not finish. Why? The installation of deps was using Playwright to also install browsers. That triggered a bug in Node where it caused the extraction of those archives to stall. We had **no escape hatch** to any of that. Even if an escape hatch exists, it has to be found and fixed on-the-fly.

Ultimately, we need to build an alternative in a composite action. That way it runs in the node context of the caller and not in its own isolated process distinct from it. Giving the callers more control over what works for them.

The old decisions don't hold weight for making both contextually aware and idempotent runners. We need a mix of both. While doing something as a Node action is highly idempotent, it also isolates context too much and can cause issues. Which is the last thing we need during a wide release push across multiple repositories. If a problem is detected it needs to be fixable.

By giving a set of rules like this, too much emphasis is placed on alignment and not enough on making the best engineering choice for a given situation.

README.md
```diff
@@ -1,17 +1,266 @@
...
+
+| If the action mainly…                                        | Use           |
+| ------------------------------------------------------------ | ------------- |
+| calls other actions / `gh` / `git` and passes values through | **Composite** |
+| runs a few shell lines with no parsing                       | **Composite** |
+| queries the GitHub API with custom GraphQL/REST              | **Node (TS)** |
```
Garbee: The GH API calls are trivial with the GitHub CLI to be done in a reasonable and maintainable way from a shell script. So I don't see why this would be a rule to follow at all for the context to use. (Sometimes, the CLI even does things natively that we don't need a deeper direct API call for.)

README.md
```diff
@@ -1,17 +1,266 @@
...
+| parses, transforms, or validates structured data             | **Node (TS)** |
```
Garbee: What exact kind of "structured data"? There is a lot and we have tools like `jq` and `yq` for JSON and YAML editing/validation/transformation. So this gets more nuanced than it seems on the surface to make the best choice.

README.md
```diff
@@ -1,17 +1,266 @@
...
+| has conditionals/policy logic you'd want to unit-test        | **Node (TS)** |
```
Garbee: We can unit test all kinds of things in all kinds of ways. Shell script? Use [bats](https://bats-core.readthedocs.io/en/stable/index.html).

I don't see why the desire to test is locked to a language. We should always find ways to validate the behavior of what we are doing.

README.md
```diff
@@ -1,17 +1,266 @@
...
+
+Rule of thumb: **YAML for wiring, TypeScript for logic.** When a composite action's `run:` blocks
+start growing `if`/loops/parsing, that logic belongs in a Node action (often a small one the
```
Garbee: If something is growing, we can break down scripts into smaller scripts focusing on certain units of work within the whole. I don't see why we'd lock a decision down to purely script size.

A big script, whether it is Bash, Node, or anything, is always a maintenance burden. The goal should be to find the right layer to abstract and do that to ease maintenance work.

README.md
```diff
@@ -1,17 +1,266 @@
...
+    default: '188'
+outputs:
+  version:
+    description: 'The extracted version'
+```
+
+### The Node action source pattern
+
+Keep `index.ts` a one-liner that injects dependencies into `run`, so `run` and the helpers stay
+pure and testable (no hidden globals):
```
Garbee: This doesn't make any sense. Maybe back in the old days this was necessary. But now with ESM and Node's baked in module mocking, we can stub out imports easily. So forcing a `run` function separated into another file makes less and less sense. It is abstraction for the sake of it over providing meaningful value in a modern system.

Is there a clear example to show why a separate run script is still needed to enable testing pathways?

README.md
```diff
@@ -1,17 +1,266 @@
...
+```typescript
+core.info(`Issue state: "${issue.state}", closed_at: "${issue.closed_at}"`)
+if (!issue.closed_at) {
+  core.info(`Issue #${issueNumber} is not closed — nothing to do`)
+  return
+}
+core.info(`Updating "${fieldName}" to ${dateString}`)
+```
+
+**Never log sensitive data.** Tokens, secrets, credentials, and full auth headers must never be
```
Garbee, line 253: This should probably be a level 4 sub-heading. And also mention in it that `core.setSecret` exists as a pragmatic way to do `::add-mask::` commands. Ref: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands?apiVersion=2026-03-10&versionId=free-pro-team%40latest&productId=rest#example-creating-an-annotation-for-an-error

This can be used to register new values that exist outside of secrets (Perhaps things auto-generated that we don't want leaked either) to help ensure they don't make it into logs either.

Garbee (dismissed):
We need to make sure what is updated is not just updated based on "current patterns" from a legacy way of working. Let's make sure they align correctly to the expectations we need going forward. Particularly since the ESM upgrade is done which brings in some benefits on making things easier to test without creating miniscule files like an `index.ts` to just import and call a run function.

README.md
```diff
@@ -1,17 +1,266 @@
 # axe-api-team-public
 
-Shared configuration, settings, and actions for the Axe API Team
+Shared configuration, settings, and reusable GitHub Actions for the Axe API Team.
 
-## Directory Structure
+This repository is the team's central library of GitHub Actions. Other repos consume these
+actions by referencing them at `dequelabs/axe-api-team-public/.github/actions/<action>@main`,
+and inherit shared repo settings via [Probot Settings](#probot-settings).
```
Zidious, line 7: Do we even use Probot?

README.md
```diff
@@ -1,17 +1,266 @@
...
+```yaml
+runs:
+  using: 'composite'
+  steps:
+    - uses: dequelabs/axe-api-team-public/.github/actions/generate-commit-list-v1@main
+      id: get-commit-list
+      with: { tag: ${{ steps.get-previous-tag.outputs.tag }} }
+    - uses: dequelabs/axe-api-team-public/.github/actions/label-and-move-released-issues-v1@main
+      with: { commit-list: ${{ steps.get-commit-list.outputs.commit-list }}, token: ${{ inputs.token }} }
+```
```
Zidious, line 68: Nit - I think linking to working examples and to the official GitHub docs for composite actions is sufficient enough that we don't need these inline.

Zidious (dismissed):
Modulo the comments left by @Garbee

README.md
Garbee: Looks like some things might. I'm honestly not entirely against keeping it mentioned until we sort it out further. Since if it is still wired up and usable, it at least helps with some of our config and tracking it. I wish it did rulesets config, but it seems not to. So, that's why we'd need to look elsewhere. Or we just do "Protobot for what it can do, but track ruleset configs somehow elsewhere." (Ocarina team repo has some helper scripts for rolling ours out and doing updates.)

I do think anything that does have Probot in use, should have a ruleset requiring an admin of that team approve the changes. That way it isn't just anyone with push outright, changes still need admin approval.

README.md
```diff
@@ -1,17 +1,266 @@
...
+start growing `if`/loops/parsing, that logic belongs in a Node action (often a small one the
+composite then calls).
+
+## Creating a new action
+
+### Directory layout & naming
+
+Each action lives in its own directory under `.github/actions/`, named `<action-name>-vN` where
+`N` is the major version. Breaking changes ship as a **new directory** (`...-v2`) so existing
+callers pinned to `...-v1@main` keep working — we never branch per action.
```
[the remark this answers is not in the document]
maksym-shynkarenko: `.github/actions/cancel-if-release-pr-exists` is the old action. All new actions should have versioning for future use

README.md
maksym-shynkarenko: will extend this section

README.md
maksym-shynkarenko: You're right that `core/github` isn't the test mechanism — the suite stubs real deps with `mock.module`, so passing them as params is stylistic, not required.
The one thing the `index.ts/run.ts` split still buys us: `index.ts` calls `run()` at module top-level, so it executes on import — keeping the logic in `run.ts` lets a test `await import('./run.ts')` without running the action (e.g. `generate-commit-list-v1/src/run.test.ts`).
I will reword the section to describe the actual `mock.module` approach.

README.md
maksym-shynkarenko: [it's used](https://github.com/dequelabs/axe-api-team-public/blob/main/.github/settings.yml)

README.md
maksym-shynkarenko: this is just an example of our actions in different use-cases, not the rule on how to use. I will add a link to the github doc as well

README.md
Zidious: The settings file exists yes, but it's not _used_ in any repo is what I was getting at. It doesn't block this PR per-say just something to consider removing if you're updating the docs anyway.

Zidious (dismissed):

maksym-shynkarenko:
@Garbee 
could you re-review the PR again?

README.md
Garbee: I don't think my point got across clearly. I am saying this abstraction of `run` separately does not buy us anything. In a correctly structured test to modern handling, you setup your mock imports before import anyways. So you always test exactly what is needed. Meaning if the import triggers the execution... That is fine, it will still just work.

So having this hard decision that we separate codified into the doc is an anti-pattern. It does not serve any functional value compared to just writing the code how you normally would.

Our functional code should always strive for direct simplicity. The test setup and apparatus will follow. This is putting some specific test setup before the functioning code.

Garbee (dismissed):
Overall looks good. Just need to chase up whatever is causing this "you must separate `run` functions" is coming from. Is that from some situation that forces it to be the only way we can do things? Or is that just the pattern we've had thus far therefore everyone is too timid to change things?

As far as I am aware, there is absolutely no functional situation that requires us to do things that way. It is purely a personal preference on how the testing should be done. So we shouldn't require one way. Let it be open.

We need to stop documenting the current implementations so heavily. We can all see that in reading the code. The problem is, as patterns evolve and new techniques exist, the docs end up driving things based on stale understandings.

maksym-shynkarenko:
> We need to stop documenting the current implementations so heavily. We can all see that in reading the code. 

@Garbee 
this is the idea of this PR - update the readme file so new contributors can understand the selected approaches in this repo and follow them

README.md
maksym-shynkarenko: Checked the suites to be sure. It's uniform across all 19 Node actions: index.ts is 4–8 lines that only call run(core, github), and each run.test.ts imports run once, then invokes it 8–16 times — one per case — with a fresh injected { getInput, setFailed, … } stub and fresh spies; static deps like @actions/exec are mock.module'd.

You're right on both counts: "so the import doesn't execute the action" wasn't the real justification, and this shouldn't be a hard rule. But the split isn't zero-value. The entry and the unit-under-test have contradictory requirements: action.yml sets main: dist/index.js, so the entry must auto-execute on load, while a test needs to import the logic without running it and call it many times. One module can't be both auto-executing-on-load and importable-without-side-effects, so it's two modules — index.ts (auto-run) and run.ts (pure handler). That's separating two responsibilities, not abstraction for its own sake.

[Reworded the doc](https://github.com/dequelabs/axe-api-team-public/commit/1a6750d670fe1388c0c44cdf50547d982591486f) to describe this as the repo convention (thin entry + exported handler for injected-deps testing), not a mandate, and dropped the import-side-effect reasoning.

Garbee (approved):

PR #422, opened 2026-07-28
feat(semantic-pr-footer): accept QA notes as a trailing section

scottmries:
Lets QA notes be written as a trailing markdown section instead of only as an
inline last line.

## Why

`run.ts` reads a single line — the last non-empty one — and `isValidFooter`
requires a keyword followed by a suffix on that same line. So this fails:

```markdown
This PR does some things.

## QA Notes

Build the binary, then confirm it starts and writes a report.
```

The last line is the prose, not the heading; and even if the heading were last,
`^qa notes:? ` would reject it because of the leading `#`s and the missing
same-line suffix. In practice that pushes everyone toward a one-line footer even
when the QA plan is several sentences, and I have seen PR bodies end up with a
`## QA Notes` section for humans *plus* a redundant `QA notes: see above` line to
satisfy the check.

## What changes

**`isValidFooter` strips markdown before matching**, so a heading or emphasised
footer is recognised — `## QA Notes: verify X`, `**No QA required**`.

Only *leading* whitespace is trimmed. That detail matters: the prefix rule requires
whitespace after the keyword (`Closes:` alone is invalid, and there is a test for
it), so trimming the end would have made every prefixed footer fail. I broke all 28
existing cases on the first attempt exactly this way.

**`run.ts` falls back to the last markdown heading** when the final line is not
itself a valid footer, which is what allows the detail to live beneath the heading.

## Deliberately kept strict

The looser rule could easily become "a heading exists somewhere", which would check
nothing. So:

- **The section must have content.** A bare `## QA Notes` with nothing beneath it
  fails — a heading promising QA notes that never arrive is not a footer.
- **Only `qa notes` may stand alone.** Issue-linking keywords still need their
  reference on the same line, so `## Closes` with `#123` underneath fails.
- **Nothing previously accepted changes.** The single-line path runs first and is
  untouched; the section check is a fallback after it fails.

## Tests

54 passing, with `isValidFooter.ts` and `run.ts` both at 100% line, branch and
function coverage. New coverage: heading and emphasis forms; bare `qa notes` accepted only
as a section heading and rejected otherwise; issue-linking keywords rejected as bare
headings; trailing-section extraction with and without content; an unrelated
trailing section rejected; and a single-line footer still taking precedence when
both are present.

`dist` is rebuilt, since the action runs `dist/index.js`.

## Rollout note

Every consuming repo pins this action at `@main`, so merging takes effect
everywhere immediately. The change is additive by construction — no previously
valid footer becomes invalid — which is why I structured it as a fallback rather
than a rewrite of the matching rules.

No QA needed: CI configuration only, and this PR's own footer exercises the unchanged inline path while the new section path is covered by unit tests.

### DIFF
--- a/.github/actions/semantic-pr-footer-v1/src/isValidFooter.test.ts
+++ b/.github/actions/semantic-pr-footer-v1/src/isValidFooter.test.ts
@@ -55,4 +55,40 @@ describe('isValidFooter', () => {
   it('ignores case', () => {
     assert.ok(isValidFooter('closes: '))
   })
+  describe('markdown', () => {
+    it('accepts a heading with a suffix on the same line', () => {
+      assert.ok(isValidFooter('## QA Notes: verify the thing'))
+    })
+
+    it('accepts a standalone footer written as a heading', () => {
+      assert.ok(isValidFooter('### No QA required'))
+    })
+
+    it('accepts an emphasised standalone footer', () => {
+      assert.ok(isValidFooter('**No QA needed**'))
+    })
+  })
+
+  describe('given a section heading', () => {
+    it('accepts a bare "QA notes" keyword', () => {
+      assert.ok(isValidFooter('## QA Notes', true))
+      assert.ok(isValidFooter('## QA notes:', true))
+      assert.ok(isValidFooter('QA notes', true))
+    })
+
+    it('still rejects issue-linking keywords, which need a reference', () => {
+      assert.ok(!isValidFooter('## Closes', true))
+      assert.ok(!isValidFooter('## Fixes', true))
+      assert.ok(!isValidFooter('## Refs', true))
+    })
+
+    it('rejects an unrelated heading', () => {
+      assert.ok(!isValidFooter('## Testing', true))
+    })
+  })
+
+  it('rejects a bare "QA notes" keyword outside a section heading', () => {
+    assert.ok(!isValidFooter('QA notes'))
+    assert.ok(!isValidFooter('## QA notes'))
+  })
 })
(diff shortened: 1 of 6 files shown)


.github/actions/semantic-pr-footer-v1/src/isValidFooter.ts
```diff
@@ -20,11 +20,54 @@ const validFooterPrefixes = [
...
+ * of an ATX heading, and the emphasis characters people wrap keywords in
+ * (`**No QA required**`).
+ *
+ * Only leading whitespace is removed. The prefix rule requires whitespace after
+ * the keyword — "Closes:" alone is invalid — so trimming the end here would make
+ * every prefixed footer fail.
+ */
+function stripMarkdown(footer: string): string {
+  return footer
+    .replace(/^\s{0,3}#{1,6}\s*/, '')
```
straker, line 50: Should add a note about why up to 3 whitespace characters and between 1 and 6 `#` (I had to go figure it out that 4 whitespace characters are turned into a block code rather than a heading).

.github/actions/semantic-pr-footer-v1/src/isValidFooter.test.ts
```diff
@@ -55,4 +55,40 @@ describe('isValidFooter', () => {
...
+      assert.ok(!isValidFooter('## Fixes', true))
+      assert.ok(!isValidFooter('## Refs', true))
+    })
+
+    it('rejects an unrelated heading', () => {
+      assert.ok(!isValidFooter('## Testing', true))
+    })
+  })
+
+  it('rejects a bare "QA notes" keyword outside a section heading', () => {
```
straker, line 90: should have tests for headings with whitespace infront (especially the cutoff from 3 to 4 whitespaces)

Garbee (changes requested):
The QA Notes when provided are just a section of content in the body. They should have no impact on the footer. The only thing about QA that impacts the footer is the `No QA Required` bit.

